PR #29, opened 2022-12-15
Add notes to explain how to change the recurring scans

Bezbran:
## Describe your changes
We want to indoduce the user how to operate the recurring kubescape/image scans from the cloud.
## Screenshots
image

### DIFF
--- a/.github/workflows/helm-chart-release.yaml
+++ b/.github/workflows/helm-chart-release.yaml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   release:
-    if: ${{ github.repository == 'kubescape/kubescape' }} # run only in KS org
+    if: ${{ github.repository == 'kubescape/helm-charts' }} # run only in KS org
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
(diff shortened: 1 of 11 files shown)


.github/workflows/helm-chart-release.yaml
```diff
@@ -10,6 +10,7 @@ on:
 
 jobs:
   release:
+    if: ${{ github.repository == 'kubescape/kubescape' }} # run only in KS org
```
dwertent, line 13:
```suggestion
    if: ${{ github.repository == 'kubescape/helm-charts' }} # run only in KS org
```

It will be better if we check only the org name.

dwertent (changes requested):

.github/workflows/helm-chart-release.yaml
```diff
@@ -10,6 +10,7 @@ on:
 
 jobs:
   release:
+    if: ${{ github.repository == 'kubescape/helm-charts' }} # run only in KS org
```
dwertent, line 13:
```suggestion
    if: ${{ github.repository_owner == 'kubescape' }} # run only in KS org
```

dwertent (approved):

PR #35, opened 2022-12-22
fix notes config-scanning URL

Bezbran:
## Describe your changes
fix notes config-scanning URL
## Screenshots - If Any (Optional)
image


.github/workflows/03-helm-release.yaml
```diff
@@ -0,0 +1,48 @@
...
+          git config user.name "$GITHUB_ACTOR"
+          git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
+          
+      - name: git checkout to new commit
+        run: git checkout ${{ inputs.COMMIT_REF }}
+
+      - name: Install Helm
+        uses: azure/setup-helm@v3
+        # with:
+        #   version: v3.4.0
```
dwertent, line 43: @matanshk Why did you remove this?

.github/workflows/03-helm-release.yaml
```diff
@@ -0,0 +1,48 @@
+name: 03-Helm chart release
+
+on:
```
dwertent, line 3: We should ignore paths:
```
    paths:
      - 'charts/**.yaml'
      - 'charts/**.tpl'
```
cc: @matanshk

matanshk: @dwertent 
please talk with me, I want to understand why

dwertent: Because we do not want the workflow to run when changes are made to these paths

matanshk: The workflow will be triggered only by the in-cluster component or by triggering it manually, it does not have any other triggers (like push); so it will never run if we are changing something in this paths anyway

.github/workflows/03-helm-release.yaml
dwertent: yes

.github/workflows/03-helm-release.yaml
dwertent: We cannot trigger this workflow when updating the helm charts manually? What if I add a field?

dwertent (approved):

PR #50, opened 2023-01-20
adding otel-collector

matthyx:


matthyx:
reopening from fork...

PR #52, opened 2023-01-23
Add CICD docs

matanshk:


dwertent (approved):
@matanshk  We might want to move this doc to a different directory, but for now it is good enough.
What about the draw.io flow?

PR #54, opened 2023-01-31
Allow to include and exclude namespaces.

cgroschupp:
## Overview
Allow to include and exclude namespaces.

## How to Test
```sh
helm template -n kubescape  kubescape-prometheus-integrator ./charts/kubescape-prometheus-integrator --set "includeNamespaces[0]=ns1" --set "includeNamespaces[1]=ns2" --set "excludeNamespaces[0]=ns3"  --set "excludeNamespaces[1]=ns4" 
```

### DIFF
--- a/charts/kubescape-prometheus-integrator/values.yaml
+++ b/charts/kubescape-prometheus-integrator/values.yaml
@@ -69,6 +69,12 @@ kubescape:
   # -- submit results to the Kubescape cloud: https://cloud.armosec.io/
   submit: false
 
+  # -- list of namespaces to include
+  includeNamespaces: []
+
+  # -- List of namespaces to exclude
+  excludeNamespaces: []
+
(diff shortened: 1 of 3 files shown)


charts/kubescape-prometheus-integrator/values.yaml
```diff
@@ -7,6 +7,12 @@ createKubescapeServiceAccount: true # TODO: move to kubescape
 # -- set the image pull secrets for private registry support 
 imagePullSecrets: ""
 
+# -- list of namespaces to include
```
dwertent, line 10: @cgroschupp This value should be in the Kubescape microservice scope since it applies only to the KS microservice

dwertent (changes requested):
Thank you for your contribution :)

dwertent (approved):
Thank you :)

PR #64, opened 2023-02-08
update otel installation instructions

dwertent:
Signed-off-by: David Wertenteil <dwertent@armosec.io>

## Overview
I added the OTEL instruction so OTEL can be added inline

### DIFF
--- a/charts/kubescape-cloud-operator/values.yaml
+++ b/charts/kubescape-cloud-operator/values.yaml
@@ -120,7 +120,7 @@ kubescape:
   image:
     # -- source code: https://github.com/kubescape/kubescape/tree/master/httphandler (public repo)
     repository: quay.io/kubescape/kubescape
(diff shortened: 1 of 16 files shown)


charts/kubescape-cloud-operator/values.yaml
```diff
@@ -424,6 +424,8 @@ otelCollector:
     host: ""
     port: 4317
     insecure: true
+    headers:
```
matthyx, line 427: these are only required when submitting to uptrace, I think it's better to not include them in `values.yaml`

dwertent: Noted. I [removed](https://github.com/kubescape/helm-charts/pull/64/commits/a7fffb2bc147473c76de24f3015595b4cbdeafe0) the headers from the example

matthyx (approved):

PR #66, opened 2023-02-09
add interval and scrapeTimeout to values.yaml

HomMarkHunt:
## Overview

I feel that using Prometheus integration makes modifying some Service Monitor properties more convenient.
Specifically, I'd like to make changes so that the interval and scrapeTimeout can be set:
1. Change `interval` to allow scans to be performed at any desired timing.
1. Modify the `scrapeTimeout` value to ensure that scans do not fail even if they exceed `100s` by executing scans with the minimum CPU limit."


dwertent (approved):
Thanks :)

PR #95, opened 2023-02-28
fix(host-scanner): remove hostNetwork and hostPort parameters

alegrey91:
## Overview
This fix removes the hostNetwork and hostPort parameters from the host-scanner deployment, since we don't need them anymore.


matthyx:
I am not sure we still use this one... I thought we always use the internal one? @dwertent ?

dwertent:
@alegrey91 Please update the kubescape image version...

dwertent:
@matthyx You might be right, I'm checking this.

dwertent (approved):

dwertent (approved):

PR #149, opened 2023-05-02
fix: move host-scanner namespace

alegrey91:
## Overview
In this PR we are moving the previous dedicated **host-scanner**'s namespace (`kubescape-host-scanner`), to the current in use for **kubescape** operator (`kubescape`).

## Additional Information

This is useful to avoid communication issues between namespaces when a policy engine is installed on the cluster.

## How to Test

Build kubescape by your own from this branch and run it against a cluster with the following command: `./kubescape scan --enable-host-scan`.
Check if host-scanner is deployed in kubescape namespace.

## Related Issues/PRs:
This is related to this other PR in **kubescape**:
https://github.com/kubescape/kubescape/pull/1217

- [ ] I have commented on my code, particularly in hard-to-understand areas

### DIFF
--- a/charts/kubescape-cloud-operator/assets/host-scanner-definition.yaml
+++ b/charts/kubescape-cloud-operator/assets/host-scanner-definition.yaml
@@ -7,13 +7,13 @@ metadata:
     k8s-app: kubescape-host-scanner
     kubernetes.io/metadata.name: kubescape-host-scanner
     tier: kubescape-host-scanner-control-plane
-  name: kubescape-host-scanner
+  name: kubescape
 ---
 apiVersion: apps/v1
 kind: DaemonSet
 metadata:
   name: host-scanner
-  namespace: kubescape-host-scanner
+  namespace: kubescape
   labels:
     app: host-scanner
     k8s-app: kubescape-host-scanner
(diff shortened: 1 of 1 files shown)


dwertent:
Here the same, we will first need to fix the system tests.
Also, did you tests this chart locally? Did you make sure Kubscape does not delete the `kubescape` namespace as part of the cleanup of the host scanner?

alegrey91:
> Here the same, we will first need to fix the system tests. Also, did you tests this chart locally? Did you make sure Kubscape does not delete the `kubescape` namespace as part of the cleanup of the host scanner?

I tested it locally, but I didn't noticed about that.
Anyway, clean up should be safe, since the teardown function here:
https://github.com/kubescape/kubescape/blob/7feea434215b68220551fc22c0da92e651d88ec4/core/pkg/hostsensorutils/hostsensordeploy.go#L304
checks if the namespace was already present before the `host-scanner` was installed.
It was a fix that we did a couple of months ago.

dwertent:
I have a hot-fix release waiting. I'll merge this once we are done with the hot-fix

dwertent (approved):

PR #168, opened 2023-05-26
remove etcd, reuse pvc, don't run as root

matthyx:


dwertent:
@matthyx Please pull the github actions from the relevancy branch so the tests will run (The relevancy tests will probably fail because the watches are not implemented yet).

matthyx:
> @matthyx Please pull the github actions from the relevancy branch so the tests will run (The relevancy tests will probably fail because the watches are not implemented yet).

rebased

dwertent:
We merged to master

PR #171, opened 2023-05-28
Add proxy support

Bezbran:
## Overview
This PR adds the ability to configure a proxy to be in use for all Helm components.
By adding the following to the `helm upgrade --install ...` command: `--set global.httpsProxy='http(s)://x.y.z'`, all the egress traffic from Helm components to external HTTPS services will be tunneled to the proxy.

In the case of a proxy that has its own TLS certificate, one needs to create a secret containing the certificate/rootCA of the certificate by running the following command:
```
kubectl -n kubescape create secret generic kubescape-proxy-tls-certificate --from-file=proxy.crt=org-proxy.crt
```
Where `org-proxy.crt` is the certificate/ RootCA of the certificate.
and after that, they should run the following command for all the Helm deployments/statefulsets (operator, kubescape, kubevuln, otel-collector, gateway):
```
kubectl -n kubescape patch [deployment] [sts] XYZ -p '{
  "spec": {
    "template": {
      "spec": {
        "volumes": [
          {
            "name": "kubescape-proxy-tls-certificate",
            "secret": {
              "secretName": "kubescape-proxy-tls-certificate"              
            }
          }
        ],
        "containers": [
          {
            "name": "XYZ",
            "volumeMounts": [
              {
                "name": "kubescape-proxy-tls-certificate",
                "mountPath": "/etc/ssl/certs/proxy.crt",
                "subPath": "proxy.crt"
              }
            ]
          }
        ]
      }
    }
  }
}'
```


Bezbran:
should do "squash and merge" ,so we won't get any GitGuardian Security Checks errors

dwertent:
Merged #173 instead

PR #178, opened 2023-05-31
add support for offline grype db

matthyx:

### DIFF
--- a/labs/kubescape-relevancy/values.yaml
+++ b/labs/kubescape-relevancy/values.yaml
@@ -297,7 +297,7 @@ kubevuln:
   image:
     # -- source code: https://github.com/kubescape/kubevuln
     repository: quay.io/kubescape/kubevuln
-    tag: v0.2.86
+    tag: v0.2.87
     pullPolicy: Always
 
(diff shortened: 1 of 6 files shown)


labs/kubescape-relevancy/values.yaml
```diff
@@ -296,8 +296,8 @@ kubevuln:
 
   image:
     # -- source code: https://github.com/kubescape/kubevuln
-    repository: quay.io/kubescape/kubevuln
```
dwertent, line 299: Maybe we should first build the new image

dwertent (changes requested):

labs/kubescape-relevancy/values.yaml
matthyx: yes, this is for @rcohencyberarmor to run it with system tests

matthyx: @dwertent done, and disabled by default

dwertent:
@matthyx can you please bump the helm chart version?

matthyx:
done, sorry

PR #184, opened 2023-06-01
add storage and relevancy

matthyx:


.github/workflows/relevancy-e2e-test.yaml
```diff
@@ -32,7 +32,15 @@ jobs:
               scan_compliance_score,
               vulnerability_scanning_cve_exceptions,
               vulnerability_scanning_test_public_registry_connectivity_by_backend,
-              vulnerability_scanning_test_public_registry_connectivity_excluded_by_backend
+              vulnerability_scanning_test_public_registry_connectivity_excluded_by_backend,
+              relevantCVEs,
```
dwertent, line 36: @rcohencyberarmor Please advise regarding the correct tests that should run when the storage is installed and the nodeAgent is not.

charts/kubescape-cloud-operator/templates/kubevuln/deployment.yaml
```diff
@@ -44,6 +44,9 @@ spec:
       imagePullSecrets:
       - name: {{ toYaml .Values.imagePullSecrets }}
       {{- end }}
+      securityContext:
```
dwertent, line 47: Was this committed intentionally?

charts/kubescape-cloud-operator/values.yaml
```diff
@@ -308,7 +308,7 @@ kubevuln:
   image:
     # -- source code: https://github.com/kubescape/kubevuln
     repository: quay.io/kubescape/kubevuln
-    tag: v0.2.86
+    tag: v0.2.88
```
dwertent, line 311:
```suggestion
    tag: v0.2.89
```

charts/kubescape-cloud-operator/values.yaml
```diff
@@ -328,7 +328,7 @@ kubevuln:
        #               https://github.com/kubescape/kubescape/issues/389
        ephemeral-storage: 5Gi
     limits:
-       cpu: 1000m
+       cpu: 500m
```
dwertent, line 331: Please revert the CPU limit decrease

charts/kubescape-cloud-operator/values.yaml
```diff
@@ -516,3 +516,43 @@ otelCollector:
...
+    deployment:
+      labels:
+        app.kubernetes.io/name: "aggregated-apiserver"
+        app.kubernetes.io/component: "apiserver"
+        app.kubernetes.io/part-of: "kubescape-storage"
+
+      replicaCount: 1
+      image:
+        pullPolicy: "IfNotPresent"
+        repository: "quay.io/matthiasb_1/storage"
```
dwertent, line 538: This image should be published to the Kubescape org.
Also, it should support `arm64` and `amd64`.

dwertent (changes requested):
I see we have in many places the `.Values.kubescape.submit` condition.
Since we will be storing the results locally, there is no need of denying the Kubevuln to run if the submit is set to `false`.
We should define the different possible behaviors and fix the dependencies accordingly.

charts/kubescape-cloud-operator/templates/kubevuln/deployment.yaml
matthyx: yes, it is needed to fix the pv permissions when we mount it

charts/kubescape-cloud-operator/values.yaml
matthyx: will work on it with @vladklokun

matthyx:
> I see we have in many places the `.Values.kubescape.submit` condition. Since we will be storing the results locally, there is no need of denying the Kubevuln to run if the submit is set to `false`. We should define the different possible behaviors and fix the dependencies accordingly.

should be much better now

charts/kubescape-cloud-operator/values.yaml
vladklokun: Yeah, that’s going to happen only when we create proper CI for Storage.

dwertent (changes requested):

charts/kubescape-cloud-operator/templates/kubescape-storage/tests/test-connection.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
```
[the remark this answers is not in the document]
vladklokun: It’s a Helm chart test I added some time ago that tests if the Storage Service is available. Think unit tests, but for Helm charts.

In my opinion, we should have more things like that if we want reproducible behavior in our customers’ clusters.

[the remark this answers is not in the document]
vladklokun: However we should make sure that this test runs only when the APIService is enabled and it makes sense to try and connect to it.

matthyx:
I merge this now, knowing that we still need an updated operator image.

PR #194, opened 2023-06-12
check GitVersion to detect managed clusters and activate node-agent

matthyx:


amirmalka (approved):

dwertent:
@matthyx What if I'm running an unmanaged cluster but I still want relevancy to run. How can I do that?

matthyx:
> @matthyx What if I'm running an unmanaged cluster but I still want relevancy to run. How can I do that?

you can't... should I add a forceEnable to relevancy?

dwertent:
Great!
We will merge this after we see the system tests are stable. Because we will need the system tests to run 
```
--set capabilities.relevancy=enable
```
instead of enabling the `node-agent`.

matthyx:
> Great! We will merge this after we see the system tests are stable. Because we will need the system tests to run
> 
> ```
> --set capabilities.relevancy=enable
> ```
> 
> instead of enabling the `node-agent`.

works on my PC btw 😅

charts/kubescape-cloud-operator/values.yaml
```diff
@@ -6,6 +6,9 @@ ksNamespace: kubescape
 ksLabel: kubescape
 createKubescapeServiceAccount: true # TODO: move to kubescape
 
+capabilities:
+  relevancy: detect # enable/disable/detect
```
dwertent, line 10: Default should be `disable`

dwertent (changes requested):

PR #222, opened 2023-07-04
fix function

Daniel-GrunbergerCA:
Fix `cloud_provider` definition, use its value in `check.provider` and in configmap.
Add `relevantImageVulnerabilitiesConfiguration` to configmap

### DIFF
--- a/charts/kubescape-cloud-operator/templates/_helpers.tpl
+++ b/charts/kubescape-cloud-operator/templates/_helpers.tpl
@@ -1,15 +1,15 @@
 {{/* standardize cloud provider */}}
 {{- define "cloud_provider" -}}
-  {{- if .Values.cloudProviderMetadata.cloudProviderEngine -}}
-    {{- $provider := lower .Values.cloudProviderMetadata.cloudProviderEngine -}}
-    {{- if or (contains "eks" $provider) (contains "aws" $provider) (contains "amazon" $provider) -}}
-eks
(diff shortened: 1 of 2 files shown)


charts/kubescape-cloud-operator/templates/_helpers.tpl
```diff
@@ -1,15 +1,15 @@
...
+    {{- print "eks" -}}
+  {{- else if contains "gke" .Capabilities.KubeVersion.GitVersion -}}
+    {{- print "gke" -}}
+  {{- else if contains "azmk8s.io" .Values.clusterServer -}}
+    {{- print "aks" -}}
+  {{- else -}}
+    {{- print "" -}}
   {{- end -}}
-{{- end }}
+{{- end -}}
```
matthyx, line 12: n00b question, what's the difference between `{{- end }}` and `{{- end -}}`?

Daniel-GrunbergerCA: I have updated it. It would trim the line twice

matthyx (approved):

dwertent (changes requested):
@Daniel-GrunbergerCA please remove `cloudProviderMetadata.cloudProviderEngine` from the values file and the documentation.

Daniel-GrunbergerCA:
@dwertent 
Done. I have also changed the include for the prometheus helm-chart

dwertent (approved):

PR #228, opened 2023-07-12
"Security" fw scan support

kooomix:
Adding to operator env var TRIGGER_SECURITY_FRAMEWORK.
If true (default), the operator will trigger scan for "security" fw on top of any framework scan.

### DIFF
--- a/charts/kubescape-cloud-operator/values.yaml
+++ b/charts/kubescape-cloud-operator/values.yaml
@@ -256,6 +256,8 @@ operator:
   # Additional volumeMounts to be mounted on the websocket
   volumeMounts: []
 
+  TriggerSecurityFramework: true
+
 kubevulnScheduler:
 
   ## Schedule Scan using cron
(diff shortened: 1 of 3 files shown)


yuleib (approved):

charts/kubescape-cloud-operator/values.yaml
```diff
@@ -256,6 +256,8 @@ operator:
   # Additional volumeMounts to be mounted on the websocket
   volumeMounts: []
 
+  TriggerSecurityFramework: true
```
dwertent, line 259: When will it ever be false?
Also, notice that all of the keys begin with lowercase and not uppercase:
`TriggerSecurityFramework` -> `triggerSecurityFramework`

kooomix: @dwertent 
(merged before your commented, opened a [new PR](https://github.com/kubescape/helm-charts/pull/229) with the small letter fix)

On helm installation, it should always be true.

PR #229, opened 2023-07-12
lower case minor fix

kooomix:


dwertent:
FYI I think the proper approach here is to list the default frameworks and not to have an environment variable for a single framework.

dwertent (approved):

PR #230, opened 2023-07-13
switching deployment strategy to recreate for PV

matthyx:


matthyx:
@dwertent if you wait I can bump node-agent and operator too

PR #246, opened 2023-08-03
[kubescape-prometheus-integrator]Default logger settings

kohbis:
## Overview

The logger configuration has been added to the integrator in the following Pull Request: https://github.com/kubescape/helm-charts/pull/185, but it has not been reflected in values.yaml.
If there are no default settings, developers will encounter value reference errors when updating the chart.

This is a similar fix as in Pull Request https://github.com/kubescape/helm-charts/pull/81.


dwertent:
Hi @kohbis , thank you for submitting.
We are working on a different integration with Prometheus, so this chart will be deprecated in the close few weeks. Stay tuned :)

PR #261, opened 2023-08-18
Update NOTES.txt

lokirosenberry:
## Overview

Corrected some typos.
The links resolved to 404s. I changed them to something that resolves, and is hopefully in-line with the original intent.


dwertent:
Thank you @lokirosenberry .
I think we will be replacing the notes as you see today with a message.
 
@craigbox , @matthyx , @slashben Now that we are removing the ARMO platform from the default provider, do you have any suggestion of what we should print after the helm installation? Maybe we should suggest how the user can see the summary results.

matthyx:
Ideally this text should be provided by the cloud we're connected to, and if we use it without a cloud, it could contain kubectl commands to see local scan results as CRDs (once available).

dwertent:
Thank you for opening this PR, we are working on the text we wish to present after the chart installation.

PR #262, opened 2023-08-18
add capabilities support

matthyx:


matthyx:
Should we deprecate `.Values.kubescape.submit`?

matthyx:
PR currently waiting for new operator configuration files... @dwertent

charts/kubescape-operator/templates/kubescape-scheduler/cronjob.yaml
```diff
@@ -2,7 +2,7 @@
   {{- fail "`batch/v1/CronJob not supported`" }}
 {{- end }}
 {{- $config := fromYaml (include "config" .) }}
-{{- if and .Values.kubescapeScheduler.enabled $config.kubescape.enabled }}
+{{- if and $config.gateway.enabled $config.kubescape.enabled }}
```
dwertent, line 5: gateway?

charts/kubescape-operator/templates/kubevuln-scheduler/configmap.yaml
```diff
@@ -1,5 +1,5 @@
 {{- $config := fromYaml (include "config" .) }}
-{{- if .Values.kubevulnScheduler.enabled }}
+{{- if $config.gateway.enabled }}
```
dwertent, line 2:
```
{{- if and $config.gateway.enabled $config.kubevuln.enabled }}
```
This way you enable only if KV is enabled

charts/kubescape-operator/templates/kubescape-scheduler/cronjob.yaml
matthyx: yes, because we said scheduled scans only make sense when we have Armo platform

charts/kubescape-operator/values.yaml
```diff
@@ -4,10 +4,23 @@
...
-createKubescapeServiceAccount: true # TODO: move to kubescape
 
 capabilities:
-  relevancy: detect # enable/disable/detect
+  relevancy: enable
+  configurationScan: enable
+  continuousScan: disable
+  vulnerabilityScan: enable
+  nodeScan: enable
+  networkGenerator: disable
```
dwertent, line 14:
```suggestion
  # networkGenerator: disable
```

charts/kubescape-operator/values.yaml
```diff
@@ -4,10 +4,23 @@
...
+  seccomp: disable
```
dwertent, line 15:
```suggestion
  # seccomp: disable
```

charts/kubescape-operator/values.yaml
```diff
@@ -4,10 +4,23 @@
...
+  runtimeObservability: disable
```
dwertent, line 16:
```suggestion
  # runtimeObservability: disable
```

charts/kubescape-operator/values.yaml
```diff
@@ -4,10 +4,23 @@
...
+  otel: enable
```
dwertent, line 17: OTEL should be disabled by default.
If there is a "third-party" we should enable it by default.
```suggestion
  otel: disable
```

charts/kubescape-operator/README.md
```diff
@@ -148,7 +148,6 @@ docker-compose logs uptrace
 | global.httpsProxy | string | `""` | Set https egress proxy for all components. Must supply also port.  |
```
dwertent, line 148: We need to update the capabilities here

charts/kubescape-operator/templates/_helpers.tpl
```diff
@@ -0,0 +1,55 @@
...
+    {{- print "gke" -}}
+  {{- else if contains "azmk8s.io" .Values.clusterServer -}}
+    {{- print "aks" -}}
+  {{- else -}}
+    {{- print "" -}}
+  {{- end }}
+{{- end }}
+
+{{- define "account_guid" -}}
+  {{- if .Values.kubescape.submit }}
```
dwertent, line 15: This is no longer relevant, KS will submit if it has URL and account ID, otherwise, KS wont submit

charts/kubescape-operator/templates/configs/cloudapi-configmap.yaml
```diff
@@ -1,6 +1,7 @@
-{{ template "account_guid" . }}
-{{ template "cluster_name" . }}
-kind: ConfigMap 
+{{- template "account_guid" . }}
```
dwertent, line 1: We should have a "configurations" section (similar to what you did with capabilities), in this configurations section we would determine if we are in "third-party mode". We would need an accountID, URL, and cluster name. 
This would prevent us from checking if we have clusterName etc.

dwertent (changes requested):

charts/kubescape-operator/values.yaml
matthyx: I think I will move the otel-url to `configurations` and use that for enabling the capability and its configuration

amirmalka (dismissed):

matthyx:
@amirmalka I have just bumped the `operator` after merging our fixes

PR #264, opened 2023-08-21
add unittests with snapshots

matthyx:
to be run with
```
helm unittest -u charts/kubescape-cloud-operator/
```


dwertent:
I think we should add this to the github actions so it would run when opening a PR

matthyx:
> I think we should add this to the github actions so it would run when opening a PR

yes good point, for the moment I am using it to ensure I don't mess things up with the refactoring...
Instructions for docker run: https://github.com/helm-unittest/helm-unittest#docker-usage

charts/kubescape-cloud-operator/tests/snapshot_test.yaml
```diff
@@ -0,0 +1,19 @@
+suite: Full Snapshot
+tests:
+  - it: matches the snapshot
+    asserts:
+      - matchSnapshot: {}
+    capabilities:
+      apiVersions:
+        - batch/v1/CronJob
+    set:
+      account: 9e6c0c2c-6bd0-4919-815b-55030de7c9a0
```
dwertent, line 10: I hope this is not a real account :)

matthyx: it was our 1-4-3 account, I don't think it's critical?

PR #267, opened 2023-08-25
bump images with vulnerabilities

matthyx:


charts/kubescape-cloud-operator/values.yaml
```diff
@@ -225,7 +225,7 @@ operator:
   image:
     # -- source code: https://github.com/kubescape/operator
     repository: quay.io/kubescape/operator
-    tag: v0.1.42-prerelease
+    tag: v0.1.43-prerelease
```
matthyx, line 228: we have to release a new chart for operator's CI to succeed

PR #285, opened 2023-09-18
Correct versions for Helm chart badges

craigbox:
## PR Type:
Bug fix

___
## PR Description:
This PR corrects the version numbers displayed in the Helm chart badges of the 'kubescape-operator' README file. The previous version was incorrectly showing as 1.11.0, but it has been updated to the correct version 1.15.3.

___
## PR Main Files Walkthrough:
`charts/kubescape-operator/README.md`: Updated the version numbers in the Helm chart badges from 1.11.0 to 1.15.3.

___
## User Description:
says 1.11
should say 1.15.3

Matthias says its OK


matthyx:
@dwertent we need to check why it only updated the text and not the badge

matthyx:
Fixed in https://github.com/kubescape/helm-charts/pull/287/files#diff-29301ee79c4acf85dcfdaa8326521bc7676bd383dff72b2e41a08c86b6727327R3

PR #299, opened 2023-10-06
add labels for servicemonitor

### LINKED ISSUE #298: be able to set labels for ServiceMonitor
## Overview
In my prometheus-operator configuration, I rely on the labels values in ServiceMonitor, so I need to be able to add my values to `labels`.

I think this is a popular practice and may be useful to others as well

## Solution
add a field to the template

batazor:
## Related issues/PRs: #298

Resolved #298

### DIFF
--- a/charts/kubescape-operator/README.md
+++ b/charts/kubescape-operator/README.md
@@ -155,7 +155,7 @@ docker-compose logs uptrace
 | kubescape.nodeSelector | object | `{}` | [Node selector](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/) |
 | kubescape.serviceMonitor.enabled | bool | `false` | enable/disable service monitor for prometheus (operator) integration |
 | kubescape.skipUpdateCheck | bool | `false` | skip check for a newer version |
-| kubescape.serviceMonitor.labels | `[]` | adds labels for servicemonitor |
+| kubescape.labels | `[]` | adds labels to the kubescape microservice |
 | kubescape.submit | bool | `true` | submit results to Kubescape SaaS: https://cloud.armosec.io/ |
 | kubescape.volumes | object | `[]` | Additional volumes for Kubescape |
 | kubescape.volumeMounts | object | `[]` | Additional volumeMounts for Kubescape |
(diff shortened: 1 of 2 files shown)


charts/kubescape-operator/templates/kubescape/servicemonitor.yaml
```diff
@@ -8,6 +8,9 @@ metadata:
   labels:
     app: {{ .Values.kubescape.name }}
     helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+    {{- with .Values.kubescape.serviceMonitor.customLabels }}
```
dwertent, line 11: I think we should have "labels" and not "customLabels".
What do you think?

dwertent (changes requested):

charts/kubescape-operator/templates/kubescape/servicemonitor.yaml
batazor: Yes, quite.

charts/kubescape-operator/templates/kubescape/servicemonitor.yaml
```diff
@@ -8,6 +8,9 @@ metadata:
   labels:
     app: {{ .Values.kubescape.name }}
     helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+    {{- with .Values.kubescape.serviceMonitor.labels }}
```
dwertent, line 11:
```suggestion
    {{- with .Values.kubescape.labels }}
```

charts/kubescape-operator/README.md
```diff
@@ -155,6 +155,7 @@ docker-compose logs uptrace
 | kubescape.nodeSelector | object | `{}` | [Node selector](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/) |
 | kubescape.serviceMonitor.enabled | bool | `false` | enable/disable service monitor for prometheus (operator) integration |
 | kubescape.skipUpdateCheck | bool | `false` | skip check for a newer version |
+| kubescape.serviceMonitor.labels | `[]` | adds labels for servicemonitor |
```
dwertent, line 158:
```suggestion
| kubescape.labels | `[]` | adds labels to the kubescape microservice |
```

dwertent (changes requested):

charts/kubescape-operator/templates/kubescape/servicemonitor.yaml
batazor: These are labels specifically for serviceMonitor, as they are only needed for prometheus-operator to work correctly. It seems to me that common labels will not work here

We can add separately `labels` or `commonLabels` for common cases.

batazor: @dwertent What do you think?

dwertent (approved):
Thanks :)

PR #301, opened 2023-10-15
Support Backend Access Key

amirmalka:
## PR Type:
Enhancement

___
## PR Description:
This PR introduces an access token secret for backend support. It creates a new secret to store the account and token information, and mounts this secret to various deployments. The account ID has also been removed from the configmap. The changes also include updates to unit tests and snapshots.

___
## PR Main Files Walkthrough:

`charts/kubescape-operator/templates/configs/cloud-secret.yaml`: A new Secret has been created to store the account and token information. The data is base64 encoded.
`charts/kubescape-operator/templates/configs/cloudapi-configmap.yaml`: The accountID field has been removed from the configmap.
`charts/kubescape-operator/templates/gateway/deployment.yaml`: The new secret is mounted to the gateway deployment.
`charts/kubescape-operator/templates/kollector/statefulset.yaml`: The new secret is mounted to the kollector statefulset.
`charts/kubescape-operator/templates/kubescape/deployment.yaml`: The new secret is mounted to the kubescape deployment.
`charts/kubescape-operator/templates/kubevuln/deployment.yaml`: The new secret is mounted to the kubevuln deployment.
`charts/kubescape-operator/templates/operator/deployment.yaml`: The new secret is mounted to the operator deployment.
`charts/kubescape-operator/values.yaml`: A new field for the cloudSecret has been added to the global values.
</details>

___


charts/kubescape-operator/templates/_helpers.tpl
```diff
@@ -11,6 +11,9 @@ submit: {{ $submit }}
     {{- if empty .Values.account -}}
       {{- fail "submitting is enabled but value for account is not defined: please register at https://cloud.armosec.io to get yours and re-run with  --set account=<your Guid>" }}
     {{- end -}}
+    {{- if empty .Values.token -}}
```
dwertent, line 14: Let's wait for this statement. 
I think we should not force the token in our first iteration, mainly because you are creating a dependency here by releasing the support in the BE.

charts/kubescape-operator/templates/configs/cloud-secret.yaml
```diff
@@ -0,0 +1,12 @@
...
+apiVersion: v1
+metadata:
+  name: {{ .Values.global.cloudSecret }}
+  namespace: {{ .Values.ksNamespace }}
+  labels:
+    app: {{ .Values.global.cloudSecret }}
+    tier: {{ .Values.global.namespaceTier }}
+type: Opaque
+data:
+  account-id: {{ .Values.account | default "" | b64enc }}
```
dwertent, line 11:
```suggestion
  account: {{ .Values.account | default "" | b64enc }}
```

dwertent (changes requested):
We will need to remove the `account` from the configMap. I believe you will open a separate PR for that.

charts/kubescape-operator/templates/configs/cloud-secret.yaml
dwertent: I agree, we just need to make sure the CLI can load the data from the secret instead.

charts/kubescape-operator/templates/configs/cloud-secret.yaml
```diff
@@ -0,0 +1,12 @@
+kind: Secret
+apiVersion: v1
+metadata:
+  name: {{ .Values.global.cloudSecret }}
+  namespace: {{ .Values.ksNamespace }}
+  labels:
```
dwertent, line 6: We should have a hard-coded label here so we will identify this secret based on the label and not the secret name

dwertent (approved):

charts/kubescape-operator/values.yaml
```diff
@@ -140,7 +141,7 @@ kubescape:
   image:
     # -- source code: https://github.com/kubescape/kubescape/tree/master/httphandler (public repo)
     repository: quay.io/kubescape/kubescape
-    tag: v2.9.2
+    tag: v2.9.3-prerelease
```
amirmalka, line 144: https://github.com/kubescape/kubescape/actions/runs/6603720522/job/17937137231

amirmalka:
https://github.com/kubescape/helm-charts/actions/runs/6603737702

charts/kubescape-operator/values.yaml
dwertent, line 144: @amirmalka  I think we should wait for the official release.

amirmalka: whatever your decide.. the other components also will not be released without tests passing .. so..

dwertent, line 144:
```suggestion
    tag: v3.0.0
```

dwertent (changes requested):

dwertent (approved):

PR #302, opened 2023-10-16
add argoCD application yaml

matthyx:
- revert to the Helm hook commit
- implement `syncPolicy` as described in http://www.zakariaamine.com/2023-06-18/argocd-with-helm-hooks/
- use `kubectl patch` to keep existing values in cloud config (especially `clusterData`)


ArgoCDApplication.yaml
```diff
@@ -0,0 +1,37 @@
...
+    helm:
+      valueFiles:
+        - values.yaml
+      parameters:
+        - name: account
+          value: '' # add account ID
+        - name: clusterName
+          value: '' # add cluster name
+        - name: server
+          value: 'api.armosec.io'
```
dwertent, line 23: We should not include `api.armosec.io` by default

dwertent (changes requested):

ArgoCDApplication.yaml
```diff
@@ -0,0 +1,37 @@
...
+  name: kubescape
+spec:
+  destination:
+    name: ''
+    namespace: kubescape
+    server: 'https://kubernetes.default.svc' # change to your server
+  source:
+    path: charts/kubescape-operator
+    repoURL: 'https://github.com/kubescape/helm-charts'
+    targetRevision: argo
```
matthyx, line 13:
```suggestion
    targetRevision: HEAD
```

dwertent: What is blocking you from changing this?

matthyx: I wanted to test it with Akinox this afternoon

PR #306, opened 2023-10-23
Removing privileged flag on the node-agent Pod

slashben:
## PR Type:
Refactoring

___
## PR Description:
This PR removes the `privileged: true` flag from the node-agent pod and adds specific capabilities to the security context of the pod. It also adds an annotation to allow the node-agent to run on kernels with AppArmor. The changes are based on similar modifications in the Inspektor Gadget project and have been tested on Ubuntu (minikube) and GCP with Linux 5.15.

___
## PR Main Files Walkthrough:

`charts/kubescape-operator/templates/node-agent/daemonset.yaml`: Removed the `privileged: true` flag from the security context of the node-agent pod and added specific capabilities. Also added an annotation to allow the node-agent to run on kernels with AppArmor.
`charts/kubescape-operator/values.yaml`: No significant changes, just a minor formatting adjustment.
</details>

___
## Overview

As per popular request I am removing the `privileged: true` flag on the node-agent.

It is based on [this](https://github.com/inspektor-gadget/inspektor-gadget/pull/348) PR and [this](https://github.com/inspektor-gadget/inspektor-gadget/commit/deeee9ea2c362c4c94a55bc07251ec1d828057dd) commit in inspektor gadget.

It was tested on Ubuntu (minikube) and GCP with Linux 5.15

### DIFF
--- a/charts/kubescape-operator/templates/node-agent/daemonset.yaml
+++ b/charts/kubescape-operator/templates/node-agent/daemonset.yaml
@@ -114,6 +114,7 @@ spec:
             {{- end }}
           securityContext:
             runAsUser: 0
+            privileged: {{ .Values.nodeAgent.privileged }}
             capabilities:
               add:
                 - SYS_ADMIN
--- a/charts/kubescape-operator/templates/node-agent/daemonset.yaml
+++ b/charts/kubescape-operator/templates/node-agent/daemonset.yaml
@@ -19,6 +19,8 @@ spec:
       tier: {{ .Values.global.namespaceTier }}
   template:
     metadata:
+      annotations:
+        container.apparmor.security.beta.kubernetes.io/node-agent: unconfined
       labels:
         app.kubernetes.io/name: {{ .Values.nodeAgent.name }}
         app.kubernetes.io/instance: {{ .Release.Name }}
@@ -112,7 +114,6 @@ spec:
             {{- end }}
           securityContext:
             runAsUser: 0
-            privileged: {{ .Values.nodeAgent.privileged }}
             capabilities:
               add:
                 - SYS_ADMIN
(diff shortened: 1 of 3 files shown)


charts/kubescape-operator/templates/node-agent/daemonset.yaml
```diff
@@ -19,6 +19,8 @@ spec:
       tier: {{ .Values.global.namespaceTier }}
   template:
     metadata:
+      annotations:
+        container.apparmor.security.beta.kubernetes.io/node-agent: unconfined
```
dwertent, line 23: @slashben can you please explain what is this annotation?

charts/kubescape-operator/templates/node-agent/daemonset.yaml
```diff
@@ -112,10 +114,17 @@ spec:
             {{- end }}
           securityContext:
             runAsUser: 0
-            privileged: true
```
dwertent, line 115: `privileged: {{ .Values.nodeAgent.privileged }}`

dwertent (changes requested):

charts/kubescape-operator/templates/node-agent/daemonset.yaml
dwertent: This will need to be added to the values file, by default it should be `false`

charts/kubescape-operator/templates/node-agent/daemonset.yaml
slashben: It removes the restrictions of AppArmor (on nodes that has AppArmor) from the `node-agent` container. This is needed since AppArmor blocks the mounting of `/sys/kernel/tracing` and limits access to `/proc/<pid>`. Beforehand, `privileged: true` bypassed AppArmor therefore we didn't need this. Now we do.

charts/kubescape-operator/templates/node-agent/daemonset.yaml
slashben: Done

dwertent (approved):

PR #317, opened 2023-10-29
Aks user assigned cluster support

rcohencyberarmor:
adding support in AKS cluster based on User Assigned Cluster.
for more information please refer to our documentation. (https://hub.armosec.io/docs/azure-aks)


Bezbran:
Do we need the same for ACR integration for image scan?
If so - we need to add the support to the operator and the kube-vuln as well.

dwertent:
True, @rcohencyberarmor please add the environment variables to the Kubevuln as well.
@Bezbran I think the Operator also authenticates with the registry.

Bezbran:
> True, @rcohencyberarmor please add the environment variables to the Kubevuln as well. @Bezbran I think the Operator also authenticates with the registry.

So we need it in the Operator deployment as well. The question is if we need to change the code of authentication with ACR.

dwertent:
@Bezbran I dont think we will need to change the code because the environment variables are supported by Azure.

dwertent (approved):

PR #324, opened 2023-10-31
Add account to storage & node-agent

amirmalka:
## PR Type:
Enhancement

___
## PR Description:
This PR enhances the security of the Kubescape Operator by loading the account from a secret instead of a config map in the storage and node-agent components. The changes include:
- Addition of a new volume for the secret in both the node-agent daemonset and the storage deployment.
- Mounting the secret volume to the respective pods.
- In the storage deployment, the account ID is now being set from the values file.

___
## PR Main Files Walkthrough:

`charts/kubescape-operator/templates/node-agent/daemonset.yaml`: A new volume for the secret has been added to the node-agent daemonset. This volume is then mounted to the pod at '/etc/credentials'. The secret is named according to the value of 'global.cloudSecret' from the values file.
`charts/kubescape-operator/templates/storage/deployment.yaml`: The account ID is now being set from the values file in the storage deployment. A new environment variable 'ACCOUNT_ID' has been added for this purpose.
</details>

___
## Overview

Account was loaded from config map instead of secret in the storage and node-agent components.
Not in this PR - new versions of storage & node-agent

Related PRs:
- https://github.com/kubescape/node-agent/pull/143
- https://github.com/kubescape/storage/pull/60


charts/kubescape-operator/templates/storage/deployment.yaml
```diff
@@ -48,6 +48,8 @@ spec:
           - name: "GOMEMLIMIT"
             value: "{{ .Values.storage.resources.requests.memory }}B"
           {{- if $components.otelCollector.enabled }}
+          - name: ACCOUNT_ID
+            value: "{{ .Values.account }}"
```
dwertent, line 52: It is best practice to use `valueFrom` so you can load the value from the secret

charts/kubescape-operator/templates/node-agent/daemonset.yaml
```diff
@@ -127,6 +130,9 @@ spec:
             seLinuxOptions:
               type: spc_t
           volumeMounts:
+          - name: {{ .Values.global.cloudSecret }}
```
dwertent, line 133: I believe we also need to update the node-agent image tag

dwertent (changes requested):

dwertent (approved):

PR #355, opened 2023-12-18
Enable overriding seLinux type for custom type tags

slashben:
## Overview
In different environments, privileged SELinux type tags can differ. In this PR we add capability for the user to override the most common type `scp_t` with custom tag that fits the give environment.

## Additional Information
This issue came around in Bottlerocket OS environment in AWS.

## Related issues/PRs:
https://github.com/bottlerocket-os/bottlerocket/issues/3677

### DIFF
--- a/charts/kubescape-operator/values.yaml
+++ b/charts/kubescape-operator/values.yaml
@@ -97,7 +97,7 @@ persistence:
 
 credentials:
   cloudSecret: # If left blank, the secret will be generated using the default name. Override this option if you already have a secret and wish to prevent Helm from creating the default secret on your behalf
- 
+
 global:
   httpsProxy: ""
   proxySecretFile: ""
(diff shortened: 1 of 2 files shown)


charts/kubescape-operator/values.yaml
```diff
@@ -605,6 +605,7 @@ nodeAgent:
           fieldPath: spec.nodeName
 
   privileged: false
+  seLinuxType: scp_t
```
matthyx, line 608:
```suggestion
  seLinuxType: spc_t
```
is that dyslexia?

matthyx (changes requested):

charts/kubescape-operator/values.yaml
slashben: Indeed, I will bring paper from the doctor 😉

matthyx (approved):

PR #357, opened 2023-12-20
Adding an option to set the runtime path to node agent

amitschendel:
## Overview

Encountered an issue where runc path wasn't found so Instead of setting env variable I think it's better to be part of the helm chart options.

### DIFF
--- a/charts/kubescape-operator/README.md
+++ b/charts/kubescape-operator/README.md
@@ -163,6 +163,7 @@ docker-compose logs uptrace
 | global.kubescapePsp.enabled | bool | `false` | Enable all privileges in Pod Security Policies for Kubescape namespace |
 | global.httpsProxy | string | `""` | Set https egress proxy for all components. Must supply also port.  |
 | global.proxySecretFile | string | `""` | Set proxy certificate / RootCA for all components to be used for proxy configured in global.httpsProxy |
+| global.overrideRuntimePath | string | `""` | Override the runtime path for all components |
 | credentials.cloudSecret | string | `""` | Leave it blank for the default secret. If you have an existing secret, override with the existing secret name to avoid Helm creating a default one |
 | kollector.affinity | object | `{}` | Assign custom [affinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/) rules to the StatefulSet |
 | kollector.env[0] | object | `{"name":"PRINT_REPORT","value":"false"}` | print in verbose mode (print all reported data) |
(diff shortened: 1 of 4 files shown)


matthyx:
please set to draft until ready for review
and don't forget to update unit tests before commit (see the README)

matthyx:
> I think it's better to be part of the helm chart options.

Sorry I don't see the option, you just put an empty env var at the moment

charts/kubescape-operator/values.yaml
```diff
@@ -112,6 +112,7 @@ global:
   networkPolicy:
     enabled: false
     createEgressRules: false
+  runtimePath: ""
```
matthyx, line 115:
```suggestion
  overrideRuntimePath: ""
```
and I would add some explanations to https://github.com/kubescape/helm-charts/blob/main/charts/kubescape-operator/README.md about how to find the correct runc path and where (here!) to set it

charts/kubescape-operator/README.md
```diff
@@ -163,6 +163,7 @@ docker-compose logs uptrace
 | global.kubescapePsp.enabled | bool | `false` | Enable all privileges in Pod Security Policies for Kubescape namespace |
 | global.httpsProxy | string | `""` | Set https egress proxy for all components. Must supply also port.  |
 | global.proxySecretFile | string | `""` | Set proxy certificate / RootCA for all components to be used for proxy configured in global.httpsProxy |
+| global.overrideRuntimePath | string | `""` | Override the runtime path for all components |
```
matthyx, line 166: technically it's only for node-agent

charts/kubescape-operator/README.md
```diff
@@ -1212,3 +1213,13 @@ To retrieve the `NetworkNeighbors` CRD for a workload, you can run the following
 ```bash
 kubectl -n <namespace> get networkneighbors <workload-kind>-<workload-name> -o yaml
 ```
+---
+## Common Issues
+* Error starting the container watcher - `(fanotify)`.
+
+  This error is usually caused by the `node-agent` not being able to find `runc` in any of the default paths.
+  This can be fixed by adding the path of `runc` to the global configuration [here](#values).
+  If you aren't sure where `runc` is located, you can run the following command to find it:
```
matthyx, line 1222: where should I run this command?
- ssh to node: usually not available
- node-agent pod: impossible since we're distroless
- temporary privileged pod: yes, but you should give the correct kubectl invocation

matthyx (changes requested):
we're getting there... don't give up!

charts/kubescape-operator/README.md
```diff
@@ -1212,3 +1213,17 @@ To retrieve the `NetworkNeighbors` CRD for a workload, you can run the following
...
+## Common Issues
+* Error starting the container watcher - `(fanotify)`.
+
+  This error is usually caused by the `node-agent` not being able to find `runc` in any of the default paths.
+  This can be fixed by adding the path of `runc` to the global configuration [here](#values).
+  If you aren't sure where `runc` is located, you can run the following command on the node to find it:
+  ```bash
+  find / -name runc 2>/dev/null
+  ```
+  In case you are in an environment where you can't access the node, once solution is to run a privileged pod on the node, and run the command from there, to run a privileged pod, run the following command:
```
matthyx, line 1226:
```suggestion
  In case you are in an environment where you can't access the node, one solution is to run a privileged pod on the node, and run the command from there. To create a privileged pod, run the following command:
```

matthyx (commented):
just one nit

matthyx (approved):

matthyx:
thanks @amitschendel !

PR #388, opened 2024-02-04
Adding gateway API objects to clusterrole

slashben:
To support https://github.com/kubescape/regolibrary/pull/570 adding access to Gateway API objects from Kubescape service account


matthyx (approved):

matthyx:
@slashben I've fixed unit tests... but you might want to fix the DCO too

dwertent (approved):

PR #403, opened 2024-03-01
prepare next release

matthyx:


charts/kubescape-operator/templates/synchronizer/configmap.yaml
```diff
@@ -92,6 +98,17 @@ data:
             "strategy": "copy"
           }
         ]
+      },
+      "httpEndpoint": {
+        "serverPort": "{{ .Values.synchronizer.service.targetPort }}",
+        "resources": [
+          {
+            "group": "test-ks",
```
dwertent, line 106: @Bezbran Maybe we should have a different name for the group

dwertent (approved):

PR #427, opened 2024-04-15
Adding alertCRD chart

dwertent:


charts/kubescape-operator/templates/node-agent/default-rule-binding-namespaced.yaml
```diff
@@ -0,0 +1,29 @@
+{{- if and .Values.alertCRD.installDefault .Values.alertCRD.scopeNamespaced }}
+apiVersion: kubescape.io/v1
+kind: RuntimeRuleAlertBinding
+metadata:
+  name: all-rules-default-namespace
+  namespace: default
```
Bezbran, line 6: chart NS, not default

charts/kubescape-operator/values.yaml
```diff
@@ -96,6 +96,12 @@ configurations:
   # Enable Prometheus annotations, to allow open-source Prometheus (not operator) to scrape metrics
   prometheusAnnotations: disable
 
+# installation of the alertCRD chart
+alertCRD:
+  installDefault: false # install the default CRD 
+  scopeClustered: false # it is better to have the CRDs in the cluster scope
```
Bezbran, line 102: We must ensure:
a. they are mutual exclusive
b. no 2 versions of the same CRD at a time

charts/kubescape-operator/values.yaml
```diff
@@ -461,7 +467,10 @@ nodeAgent:
     updatePeriod: 10m # duration string
     prometheusExporter: enable
     httpExporterConfig: 
-      url: http://synchronizer.kubescape.svc.cluster.local:8089/apis/v1/kubescape.io/v1/runtimealerts
+      url: http://synchronizer.svc.cluster.local:8089/apis/v1/kubescape.io/v1/runtimealerts
```
Bezbran, line 470: max alerts per minute configurations?

charts/kubescape-operator/templates/node-agent/default-rule-binding-namespaced.yaml
dwertent: Why? This means you will have the alerts only on the KS namespace. (the RB crd affects only the namespace and not other namespaces, users that do not have cluster-wide permissions will need to create similar CRDs on the namespaces they want to create alerts)

charts/kubescape-operator/values.yaml
dwertent: I [set the default](https://github.com/kubescape/node-agent/pull/257) to 100.
I'm not sure we want this number to be easily modified

charts/kubescape-operator/templates/node-agent/default-rule-binding-namespaced.yaml
Bezbran: In any case we have the namespace selector in the object.
If the RB is rab is namespaced scope is just a location of the object, but has no effect on the logic.

dwertent: I reverted this behavior for now, but we will need to think of a better solution

Bezbran (dismissed):
Just we need to make sure in node-agent the Rb is NOT required to be in the same NS as the pod

amitschendel (approved):

PR #435, opened 2024-05-08
Adding multiply

dwertent:


dwertent:
This is not meant to merge.
@bvolovat this is the branch you need for testing.
Notice, that the replicas are 3. we will probably change it to something higher

PR #445, opened 2024-05-30
add missing network policies

matthyx:


charts/kubescape-operator/values.yaml
```diff
@@ -163,6 +163,31 @@ global:
...
+              matchLabels:
+                k8s-app: kube-dns
+      - ports:
+          - port: 4317
+            protocol: TCP
+        to:
+          - podSelector:
+              matchLabels:
+                app: otel-collector
+    k8sApiEgressRules:
```
matthyx, line 184: the idea is that we list here rules to reach the api-server for well-known distributions (minikube, kind, GKE, ...) but we let users override it to their needs

charts/kubescape-operator/values.yaml
```diff
@@ -163,6 +163,26 @@ global:
   networkPolicy:
     enabled: false
     createEgressRules: false
+    apiServerIP: ""
```
matthyx, line 166: @dwertent something like that?

charts/kubescape-operator/assets/api-server-network-rules.yaml
```diff
@@ -0,0 +1,16 @@
+- ports:
+    - port: {{ .Values.global.networkPolicy.apiServerPort }}
+      protocol: TCP
+  to:
+{{- if .Values.global.networkPolicy.apiServerIP }}
```
matthyx, line 5: we can either specify the IP or use a selector - depending on the distro

slashben (approved):

slashben:
I have tested it with Squid proxy with these settings:

```yaml
networkPolicy:
  enabled: true
  createEgressRules: true
  httpsProxyIP: "10.107.10.92"
  httpsProxyPort: "3128"
server: "api.armosec.io"
account: XXXXXXXXXXXXXXXXXX
accessKey: XXXXXXXXXXXXXXXXXX
clusterName: "network-kind"
global:
  httpsProxy: "http://10.107.10.92:3128"
```

The traffic went through the proxy and network policy was enabled.

It worked

slashben (approved):

PR #448, opened 2024-06-04
Changing KDR defaults to include installation of CRD

slashben:
## Overview

1. Making the installation of runtime detection rule CRD the default behavior
2. Adding check that if the runtime detection enabled, at least one of the CRD types has to be installed

### DIFF
--- a/charts/kubescape-operator/templates/node-agent/configmap.yaml
+++ b/charts/kubescape-operator/templates/node-agent/configmap.yaml
@@ -1,6 +1,12 @@
 {{- $components := fromYaml (include "components" .) }}
 {{- $configurations := fromYaml (include "configurations" .) }}
 {{- if $components.nodeAgent.enabled }}
+{{/* validate that either alertCRD.scopeClustered or alertCRD.scopeNamespaced defined when the capability enabled */}}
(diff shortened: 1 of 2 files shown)


charts/kubescape-operator/templates/node-agent/configmap.yaml
```diff
@@ -1,6 +1,12 @@
 {{- $components := fromYaml (include "components" .) }}
 {{- $configurations := fromYaml (include "configurations" .) }}
 {{- if $components.nodeAgent.enabled }}
+{{/* validate that either alertCRD.scopeClustered or alertCRD.scopeNamespaced defined when the capability enabled */}}
```
matthyx, line 4: I try to have all the logic concentrated in _helpers.tpl - see if it makes sense to move that there

slashben: I tried!
The problem is that when I move it there, it looks like the test is not working (I does not fail templating)

matthyx: ok whatever... you know what I think about Helm

matthyx (approved):

PR #450, opened 2024-06-10
Add support for custom CA certificates in KS containers

slashben:
## Overview

Add support for custom CA certificates in Kubescape-operator deployment.

```
global.overrideDefaultCaCertificates.enabled | bool | `false` | Use to enable custom CA Certificates
global.overrideDefaultCaCertificates.caCertificates | string | `""` | Set the custom CA Certificates file in all container
```

### DIFF
--- a/charts/kubescape-operator/values.yaml
+++ b/charts/kubescape-operator/values.yaml
@@ -164,6 +164,9 @@ global:
     enabled: false
     createEgressRules: false
   overrideRuntimePath: ""
+  overrideDefaultCaCertificates:
+    enabled: false
+    caCertificates: ""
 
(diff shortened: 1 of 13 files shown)


charts/kubescape-operator/values.yaml
```diff
@@ -164,6 +164,9 @@ global:
     enabled: false
     createEgressRules: false
   overrideRuntimePath: ""
+  overrideDefaultCaCertificates:
+    enabled: false
```
matthyx, line 168: let's enable it in snapshot_test.yaml for the `all capabilities` scenario so we can see the effects

matthyx (approved):

PR #469, opened 2024-07-10
altered apiVersion and added required parameters in ArgoCDApplication.yaml

anirudh-hegde:
image


matthyx (approved):

matthyx:
thanks @anirudh-hegde

PR #484, opened 2024-08-21
Assign appropriate SCC RoleBindings for Kubescape components:

DanDavidA:
- Node-agent assigned with the privileged SCC.
- All other components assigned with the nonroot SCC, following the principle of least privilege.

### DIFF
--- a/charts/kubescape-operator/values.yaml
+++ b/charts/kubescape-operator/values.yaml
@@ -187,6 +187,9 @@ global:
   overrideDefaultCaCertificates:
     enabled: false
     caCertificates: ""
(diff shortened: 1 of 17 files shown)


charts/kubescape-operator/values.yaml
```diff
@@ -187,6 +187,9 @@ global:
   overrideDefaultCaCertificates:
     enabled: false
     caCertificates: ""
+  openshift: # Openshift Security Context Constraint support
```
matthyx, line 190: this is the default, you could have avoided it, but it's fine

matthyx (approved):

PR #486, opened 2024-08-22
Standard labels

smalinet:
## Overview

This PR is about getting closer to best practices for using helm chart, especially regarding labels.

https://helm.sh/docs/chart_best_practices/labels/#standard-labels


charts/kubescape-operator/templates/gateway/deployment.yaml
```diff
@@ -25,8 +24,7 @@ spec:
     type: RollingUpdate
   selector:
     matchLabels:
-      app.kubernetes.io/name: {{ .Values.gateway.name }}
-      app.kubernetes.io/instance: {{ .Release.Name }}
```
matthyx, line 29: @smalinet it's not clear to me whether we want to remove `app.kubernetes.io/instance` from all `matchLabels` - what's the best practice?

smalinet: @matthyx These keys are dynamically generated by the `selectorLabels` in the __helpers.tpl_ file.
This is a first iteration but in my opinion the good approach would be something like:

```yaml
    matchLabels:
      {{- include "kubescape-operator.selectorLabels" . | nindent 6 }}
      app.kubernetes.io/component: {{ .Values.gateway.name }}
```

charts/kubescape-operator/templates/_helpers.tpl
```diff
@@ -1,103 +1,66 @@
...
+{{/*
+Selector labels
+*/}}
+{{- define "kubescape-operator.selectorLabels" -}}
+app.kubernetes.io/name: {{ include "kubescape-operator.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+app.kubernetes.io/component: {{ .app }}
+{{- end }}
 
+{{/*
```
smalinet, line 57: I think the template for serviceAccountName can be removed.

matthyx (approved):

PR #499, opened 2024-09-09
add permissions to application profile

jnathangreeg:


matthyx (approved):

matthyx:
@jnathangreeg the release process is different for charts... but your change will be included in the next one

PR #554, opened 2024-12-01
image pull secret generation using provided credentials

amirmalka:
## Overview

This update enhances the handling of `imagePullSecrets` by allowing users to generate an image pull secret dynamically based on provided credentials. Previously, the `imagePullSecrets` value was limited to referencing an existing secret. With this change, users can specify credentials and the chart will generate the required secret.


amirmalka:
@matthyx WDYT? I did consider "merging" the `imagePullSecrets` to a single value but end up with this...

matthyx (approved):

PR #568, opened 2024-12-12
fixed copy paste mistake for Prometheus Exporter volumes

hebestreit:
## Overview
When I added volume mounts for the kubevuln component, I also noticed differences for the Prometheus Exporter component in Argo CD. I assume this is a copy-paste error that has not yet been discovered.

### DIFF
--- a/charts/kubescape-operator/values.yaml
+++ b/charts/kubescape-operator/values.yaml
@@ -774,6 +774,11 @@ prometheusExporter:
 
   enableWorkloadMetrics: false
 
+  # Additional volumes to be mounted on the prometheus exporter
+  volumes: [ ]
+
+  # Additional volumeMounts to be mounted on the prometheus exporter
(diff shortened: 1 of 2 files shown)


charts/kubescape-operator/values.yaml
```diff
@@ -772,6 +772,12 @@ prometheusExporter:
     targetPort: 8080
     protocol: TCP
 
+  # Additional volumes to be mounted on the vulnerability scanning microservice
```
amirmalka, line 775: If we are fixing a copy-paste mistake let's also fix the comments :)

hebestreit: Good catch! Will do accordingly.

hebestreit: Comments have been updated ;-)

amirmalka (approved):

PR #573, opened 2024-12-18
allow prometheus-exporter to export metrics on workload level

hebestreit:
## Overview

With this change it's possible to set extra environment variables for the prometheus-exporter. This is desired to set `ENABLE_WORKLOAD_METRICS=true`. Additionally I've updated the image tag to the latest version.

## Additional Information

The logic has been implemented in this PR:
https://github.com/kubescape/prometheus-exporter/pull/24

~~Don't be confused by the version reduction associated with CI creating negative versions:~~
https://github.com/kubescape/prometheus-exporter/issues/26

The Docker tag `v0.2.0` has been created manually to avoid the above problem.
https://quay.io/repository/kubescape/prometheus-exporter?tab=tags

## How to Test

Create a `values.yaml` with minimal config:

```yaml
capabilities:
  prometheusExporter: enable
prometheusExporter:
  env:
    - name: ENABLE_WORKLOAD_METRICS
      value: "true"
```

Render the helm template and and check the occurrence:
```
helm template kubescape-operator kubescape-operator -f values.yaml | grep -5 "ENABLE_WORKLOAD_METRICS"
                  resource: limits.cpu
            - name: KS_LOGGER_LEVEL
              value: "info"
            - name: KS_LOGGER_NAME
              value: "zap"
            - name: ENABLE_WORKLOAD_METRICS
              value: "true"
          volumeMounts:
            - name: ks-cloud-config
              mountPath: /etc/config
              readOnly: true

```

### DIFF
--- a/charts/kubescape-operator/values.yaml
+++ b/charts/kubescape-operator/values.yaml
@@ -756,7 +756,7 @@ prometheusExporter:
   name: "prometheus-exporter"
   image:
     repository: quay.io/kubescape/prometheus-exporter
-    tag: v0.0.129
+    tag: v0.2.0
     pullPolicy: IfNotPresent
 
   resources:
@@ -771,6 +771,9 @@ prometheusExporter:
     port: 8080
     targetPort: 8080
     protocol: TCP
+
+  enableWorkloadMetrics: false
+
 
 # +++++++++++++++++++++++++++++ Upgrader ++++++++++++++++++++++++++++++++++++++++++++++++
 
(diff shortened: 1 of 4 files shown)


charts/kubescape-operator/values.yaml
```diff
@@ -772,6 +772,11 @@ prometheusExporter:
     targetPort: 8080
     protocol: TCP
 
+  env: []
```
matthyx, line 775: I would prefer to have a variable enableWorkloadMetrics that would add ENABLE_WORKLOAD_METRICS in the deployment

matthyx (changes requested):
please also enable the new feature in `snapshot_test.yaml` ("all capabilities" scenario) and update unittests

charts/kubescape-operator/values.yaml
hebestreit: Sounds good to me too. Please take a look and let me know if I've missed anything.

matthyx (approved):
awesome, thanks @hebestreit !
I'm not sure we'll release before Christmas, but definitely in the beginning of 2025

PR #584, opened 2025-01-07
fix: annotation for node-agent should use dynamic name

cyberox:
## Overview

Modified the AppArmor profile configuration to use the `nodeAgent.name` value from the Helm chart values instead of hardcoding `node-agent`. This change allows for more flexible naming of the node agent container.


amirmalka:
thanks @cyberox , please follow https://github.com/kubescape/helm-charts/tree/main/charts/kubescape-operator/tests to update the snapshot

cyberox:
@amirmalka, I updated the snapshot. Thanks for pointing me to the documentation.

amirmalka (approved):

Issue #581, opened 2025-01-05
Make the request-body.json in kubescape-scheduler configurable


Bezbran:
## Overview
Currently it is hard coded for the default recurring scan to scan all frameworks every scan.

## Problem
Users want the ability to determine what will be scanned by the recurring scan.

## Solution
1. Change the [template file](https://github.com/kubescape/helm-charts/blob/main/charts/kubescape-operator/templates/kubescape-scheduler/configmap.yaml) to take from values the request body.
2. Add the current body as default value in values file for backward compatibility.

## Alternatives
Delete the default recurring scan after installation and make a custom recurring scan as described [here](https://github.com/kubescape/operator#create-a-cronjob-that-will-repeatedly-trigger-a-kubescape-scann-according-to-a-specific-framework).

## Additional context
Issue  #[1755](https://github.com/kubescape/kubescape/issues/1755) in Kubescape repository.

matthyx:
should have been closed by https://github.com/kubescape/helm-charts/pull/590

PR #590, opened 2025-01-09
Ks recurring body configurable

Bezbran:
Solves #581


Bezbran:
We need to ignore whitespaces in JSONs in unitests

matthyx (approved):
thanks Bez!

PR #616, opened 2025-02-12
DO NOT MERGE update operator and operator-command.crd.yaml with string payload field

refaelm92:

### DIFF
--- a/charts/kubescape-operator/values.yaml
+++ b/charts/kubescape-operator/values.yaml
@@ -286,8 +286,8 @@ operator:
 
   image:
     # -- source code: https://github.com/kubescape/operator
-    repository: quay.io/kubescape/operator
-    tag: v0.2.69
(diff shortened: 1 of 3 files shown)


charts/kubescape-operator/values.yaml
```diff
@@ -286,8 +286,8 @@ operator:
 
   image:
     # -- source code: https://github.com/kubescape/operator
-    repository: quay.io/kubescape/operator
-    tag: v0.2.69
+    repository: quay.io/matthiasb_1/operator
```
dwertent, line 289: Is this an accident?

dwertent (changes requested):

amirmalka:
I think we can close this one right?

refaelm92:
> I think we can close this one right?

Yes

PR #646, opened 2025-03-12
include more resources to ignoreDifferences ArgoCD App

StefanCenusa:
## Overview

I've included some new items to the ignoreDifferences array. These resources have random generated data when helm templated causing ArgoCD to generate them at each refresh and marking the Application as not synced.


matthyx (approved):
thanks @StefanCenusa !

PR #673, opened 2025-04-09
fix http detection flag

afek854:


matthyx:
@afek854 any news on that?

matthyx (dismissed):

matthyx (approved):

PR #680, opened 2025-04-28
generate and sync filtered SBOMs with SBOMs

matthyx:


matthyx:
@kooomix we can try that in dev

kooomix (approved):

PR #686, opened 2025-05-04
Update node-agent volume mounts to specify read-only access where applicable

slashben:
This pull request introduces changes to improve the `kubescape-operator's security posture by setting appropriate `readOnly` flags for volume mounts in the `node-agent` configuration. These changes ensure that volumes are marked as read-only wherever possible, reducing the risk of unintended write operations.

### Security Enhancements:

* **`daemonset.yaml` Updates**: Added `readOnly: true` to volume mounts for `proxy-secret`, `custom-ca-certificates`, and `extra-ca-certificates` to ensure these sensitive files are mounted as read-only.

* **`values.yaml` Updates**:
  - Marked several volume mounts as `readOnly: true`, including `/host`, `/var`, `/lib/modules`, `/sys/fs/cgroup`, `/sys/fs/bpf`, and `/boot`, to restrict write access.
  - Retained write access (`readOnly: false`) for specific paths like `/run` (container runtime socket), `/sys/kernel/debug` (eBPF program loading), and `/data` and `/clamav` (emptyDir volumes), where write operations are necessary.

### DIFF
--- a/charts/kubescape-operator/templates/node-agent/daemonset.yaml
+++ b/charts/kubescape-operator/templates/node-agent/daemonset.yaml
@@ -266,17 +266,20 @@ spec:
           - name: proxy-secret
             mountPath: /etc/ssl/certs/proxy.crt
             subPath: proxy.crt
+            readOnly: true
           {{- end }}
           {{- if .Values.global.overrideDefaultCaCertificates.enabled }}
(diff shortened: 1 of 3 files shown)


matthyx (approved):
nice with comments and all 👍 
thanks @slashben !

PR #693, opened 2025-05-09
Allow setting up podLabels and podAnnotations

shivamverma182:
## Overview
We have a use case that requires deploying Kubescape in namespaces where Istio is enabled. Since Kubescape communicates with the Kubernetes API server—which is outside the service mesh—we need to exclude this traffic from mTLS. This can be achieved using specific pod annotations.

Additionally, there are several CronJobs that spawn pods with Istio sidecars. These sidecars interfere with the job lifecycle, causing them to run indefinitely. To prevent this, we can apply pod labels to disable sidecar injection for these CronJobs.

This merge request introduces support for configuring podLabels and podAnnotations.

### DIFF
--- a/charts/kubescape-operator/Chart.yaml
+++ b/charts/kubescape-operator/Chart.yaml
@@ -9,14 +9,14 @@ type: application
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
 
-version: 1.27.4
+version: 1.27.5
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
 # It is recommended to use it with quotes.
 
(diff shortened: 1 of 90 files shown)


matthyx:
hi @shivamverma182 did you try using https://github.com/kubescape/helm-charts/blob/0ec0f1150dc68c20ff84ca874421b13d52903f4d/charts/kubescape-operator/values.yaml#L21-L23 ?

shivamverma182:
Hi @matthyx, I had a look at addtional labels/annotations but different component requires annotations. Like e.g. 

1. Not all the services talk to kubeapi server, so not all services require traffic.sidecar.istio.io/excludeInboundTraffic annotation.
2. We don't want to inject sidecars to the cronjob pods since it prevents graceful completionn. If I add sidecar.istio.io/inject=false to addtionalLabels, it will get applied to all the other services and keeping them out of istio service mesh.

matthyx (approved):
thanks @shivamverma182 !

PR #694, opened 2025-05-09
Remove dependency on `submit` for enabling Kubescape scheduler pods

AdonisOkpidi:
This PR updates the `_common.tpl` template logic to decouple `kubescapeScheduler` and `kubevulnScheduler` from the [submit ](https://github.com/kubescape/helm-charts/blob/main/charts/kubescape-operator/templates/_common.tpl#L33). Previously, these pods were only enabled when submit was true, which requires an Armo Cloud account.

With this change:
- `kubescapeScheduler` is enabled based solely on `.Values.capabilities.configurationScan == "enable"`
- `kubevulnScheduler` is enabled based solely on `.Values.capabilities.vulnerabilityScan == "enable"`

This simplifies installation for users running Kubescape in self-managed mode or without an Armo Cloud integration.

### DIFF
--- a/charts/kubescape-operator/templates/_common.tpl
+++ b/charts/kubescape-operator/templates/_common.tpl
@@ -51,11 +51,11 @@ hostScanner:
 kubescape:
   enabled: {{ eq .Values.capabilities.configurationScan "enable" }}
 kubescapeScheduler:
-  enabled: {{ and $configurations.submit (eq .Values.capabilities.configurationScan "enable") }}
+  enabled: {{ eq .Values.capabilities.configurationScan "enable" }}
 kubevuln:
   enabled: {{ eq .Values.capabilities.vulnerabilityScan "enable" }}
 kubevulnScheduler:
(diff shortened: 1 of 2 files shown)


charts/kubescape-operator/templates/_common.tpl
```diff
@@ -51,11 +51,11 @@ hostScanner:
 kubescape:
   enabled: {{ eq .Values.capabilities.configurationScan "enable" }}
 kubescapeScheduler:
-  enabled: {{ and $configurations.submit (eq .Values.capabilities.configurationScan "enable") }}
+  enabled: {{ eq .Values.capabilities.configurationScan "enable" }}
```
matthyx, line 54: have you tried running a cluster like that? if I remember correctly we had issues running with both continuousScan and scheduler

matthyx (approved):

shivamverma182:
@matthyx  Can i request to create a release please

matthyx:
Yes it'll come out this week

Issue #740, opened 2025-09-03
Add support for configuring podSecurityContext and container securityContext via values.yaml


AdonisOkpidi:
## Overview
Our engineering team reported that the Kubescape operator deployment fails Kyverno PodSecurity policy validations in clusters with Pod Security admission enforced. The root cause is upstream Helm chart limitations that prevent configuring required Pod and container security context fields through values.yaml.

## Problem
The Kubescape operator deployment is failing Kyverno PodSecurity policy validations due to upstream chart limitations that prevent setting required security context fields through Helm values.

### 1. Upstream chart limitations

- The Kubescape Helm chart currently hardcodes or omits securityContext values in deployment templates.

- The values.yaml does not expose podSecurityContext or container-level securityContext options.

- This prevents users from setting required fields like runAsUser, runAsGroup, allowPrivilegeEscalation, etc. via standard Helm values.

### 2. Specific Kyverno violations observed

- require-run-as-non-root-user: runAsUser not set to >0

- require-non-root-groups: runAsGroup not set to >0

- deny-privilege-escalation: allowPrivilegeEscalation not set to false

- require-podsecurity-restricted: missing capabilities.drop, seccompProfile

- require-run-as-nonroot: runAsNonRoot not set to true
 
## Attempted configuration (ineffective)

We tried to configure security contexts in our kubescape.yaml values file, but the chart ignores them because the templates don’t reference these fields:

```
podSecurityContext:
  runAsNonRoot: true
  runAsUser: 65532
  runAsGroup: 65532
securityContext:
  allowPrivilegeEscalation: false
  capabilities:
    drop: [ALL]
  seccompProfile:
    type: RuntimeDefault
```

## Proposed solution
- Expose podSecurityContext and container-level securityContext in values.yaml.

slashben:
I'll do

AdonisOkpidi:
@slashben I have just got back now. I can see you have an open PR for this. Do you suggest testing with your branch?

matthyx:
@AdonisOkpidi did you test the branch?

AdonisOkpidi:
@matthyx still nothing yet today and as they are in India I couldn't get any response. 

[...]

PR #742, opened 2025-09-17
Add pod and container security contexts for all components

slashben:
This pull request refactors the way Kubernetes security contexts are defined in all Helm chart templates for the Kubescape operator and its components. The main improvement is to allow full customization of both pod-level and container-level `securityContext` via Helm values, replacing hardcoded defaults. This change increases flexibility for users and improves security posture by supporting best practices such as dropping all capabilities and enforcing non-root execution. Snapshot test files are updated to reflect the new, more secure defaults.

This PR is addressing #740 

**Security context templating improvements:**

* All deployment and daemonset templates now use `coalesce` to allow users to override both pod-level and container-level `securityContext` via values such as `.Values.<component>.podSecurityContext` and `.Values.<component>.securityContext`, falling back to global defaults if not set. Hardcoded fields (e.g., `runAsNonRoot`, `allowPrivilegeEscalation`, `seccompProfile`) are removed in favor of user-supplied YAML. 
**Security best practices:**

* Container-level `securityContext` now supports dropping all Linux capabilities (`capabilities: drop: [ALL]`), enforcing `readOnlyRootFilesystem`, and running as non-root by default. These changes are reflected in the updated snapshot test files.

**Component-specific enhancements:**

* The `node-agent` daemonset template now conditionally applies `appArmorProfile` if Kubernetes version is 1.29+, and allows SELinux options and privilege settings to be customized or defaulted based on values. 

**Snapshot test updates:**

* All affected snapshot test files are updated to reflect the new, configurable security context defaults, including dropped capabilities, enforced non-root execution, and group/user settings.

**Generalization and maintainability:**

* By removing hardcoded security settings and using Helm's `coalesce` function, the charts are now easier to maintain and extend, and users can more easily align deployments with organizational security policies.

### DIFF
--- a/charts/kubescape-operator/templates/grype-offline-db/deployment.yaml
+++ b/charts/kubescape-operator/templates/grype-offline-db/deployment.yaml
@@ -30,9 +30,10 @@ spec:
         {{- with .Values.grypeOfflineDB.podLabels }}{{- toYaml . | nindent 8 }}{{- end }}
         kubescape.io/tier: "core"
     spec:
+      {{- with (coalesce .Values.grypeOfflineDB.podSecurityContext .Values.global.podSecurityContext) }}
       securityContext:
-        seccompProfile:
-          type: RuntimeDefault
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       {{- if .Values.imagePullSecrets }}
       imagePullSecrets:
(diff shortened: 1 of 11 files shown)


AdonisOkpidi:
Still awaiting feedback from the team that raised this issue. I've just chased up again and will update when I receive some feedback.

PR #764, opened 2025-11-26
Generate node-agent daemonsets from node labels

matthyx:
When multipleDaemonSets.configurations is empty, lookup cluster Nodes and auto-generate configurations by grouping nodes by labels (kubescape.io/daemonset-group, node.kubernetes.io/instance-type, beta.kubernetes.io/instance-type). Use a normalized nameSuffix per group for daemonset names and add kubescape.io/daemonset-group labels to the selector and pod template.


matthyx:
pure vibe coding to test Gemini 3 in my IDE

PR #784, opened 2026-01-15
chore: adding the ability to adjust the source of busybox

### LINKED ISSUE #783: enable override of busybox image source
## Overview
Right now the busybox image for startup jitter is automatically pulled from docker with no option to override.

## Problem
If you hit the api pull limit then you can't roll out the busybox image to support the startup jitter.

## Solution
A new field in the values that allows a user to override the busybox image to use a registry of their choice. The understanding being if it's a private repo, the imagePull Secret is the same for all images.

## Alternatives
N/A

## Additional context
N/A

drew-viles:
# Overview
As a user that will deploy kubescape across a number of different environments, one problem we might hit it the docker pull limits when pulling the busybox image for the startup jitter. This PR enables a user to override the source and use an internal repo.

For example, I'm using an image pull secret for the pullthrough registry I have and it's be nice if I could use this registry for the init container too if it's enabled.

## How to Test

Install the helm chart with the startupJitter enabled and supply a custom image.

## Related issues/PRs:
Resolved #783

- [ ] New and existing unit tests pass locally with my changes

### DIFF
--- a/charts/kubescape-operator/templates/node-agent/_node-agent.tpl
+++ b/charts/kubescape-operator/templates/node-agent/_node-agent.tpl
@@ -251,7 +251,7 @@ Parameters:
 {{- define "node-agent.initContainers" -}}
 {{- if .Values.nodeAgent.startupJitterContainer.enabled }}
 - name: startup-jitter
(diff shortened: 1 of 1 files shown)


drew-viles:
@codex review

PR #793, opened 2026-02-01
add startup probe

YakirOren:
## Summary by CodeRabbit

* **Chores**
  * Updated node-agent health checks: added a startup probe and adjusted liveness/readiness behavior to improve container startup handling and deployment stability.

### DIFF
--- a/charts/kubescape-operator/templates/node-agent/_node-agent.tpl
+++ b/charts/kubescape-operator/templates/node-agent/_node-agent.tpl
@@ -186,7 +186,6 @@ Parameters:
     httpGet:
       path: /livez
       port: 7888
-    initialDelaySeconds: 60
     periodSeconds: 3
   readinessProbe:
     httpGet:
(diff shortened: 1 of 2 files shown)


charts/kubescape-operator/templates/node-agent/_node-agent.tpl
```diff
@@ -192,8 +192,15 @@ Parameters:
...
       port: 7888
-    initialDelaySeconds: 3
     periodSeconds: 3
+  startupProbe:
+    httpGet:
+      path: /readyz
+      port: 7888
+    initialDelaySeconds: 5
+    periodSeconds: 2
+    failureThreshold: 60
```
matthyx: is 2*60s enough? if that fails after 2min we'll still go crashloop

charts/kubescape-operator/templates/node-agent/_node-agent.tpl
```diff
@@ -192,8 +192,15 @@ Parameters:
     httpGet:
       path: /readyz
       port: 7888
-    initialDelaySeconds: 3
     periodSeconds: 3
+  startupProbe:
+    httpGet:
+      path: /readyz
+      port: 7888
+    initialDelaySeconds: 5
```
matthyx: useless for a startup probe I would say

PR #813, opened 2026-04-08
Enable setting admission service type, and service annotations

seppelucas:
To better support Cilium loadbalancer services in BGP clusters. Issue: #812

## Overview
Enable setting the service type for the admission controller service.
Enable settings service annotations for all services currently in values.yaml

## Additional Information
To support cilium bgp clusters where the masters are not in the clusters as nodes, needing loadbalancer services for admission controllers.

 

## Summary by CodeRabbit

* **New Features**
  * Service annotations can now be configured for kubescape, kubevuln, operator, and synchronizer; when provided they are rendered into each Service’s metadata.
  * Admission service endpoint is now configurable: service type, port, and target port can be customized via chart values, with sensible defaults included.


matthyx (approved):
thanks @seppelucas !

PR #816, opened 2026-04-09
fix(storage): allow configuring storage Service exposure

officialasishkumar:
## Overview
This updates the storage Service in the `kubescape-operator` chart so clusters that need the aggregated storage API reachable outside the cluster network can override the Service type and attach storage-specific Service annotations.

## Additional Information
The storage Service was the remaining hardcoded `ClusterIP` in the path reported in #806. The chart already exposed similar service configuration for other components, so this change keeps storage aligned without introducing a second target-port source of truth.

## How to Test
1. Run `helm unittest -u charts/kubescape-operator`.
2. Confirm the new `storage service configuration` test passes.
3. Optionally render the chart with `storage.service.type=LoadBalancer` and a storage Service annotation override to verify the generated Service manifest.

## Related issues/PRs:
* Relates #806

## Release Notes

* **New Features**
  * Storage service configuration now supports customizable service type and annotations, enabling flexible deployment options across different environments.

* **Tests**
  * Added snapshot test case for validating storage service configuration behavior.

### DIFF
--- a/charts/kubescape-operator/templates/storage/service.yaml
+++ b/charts/kubescape-operator/templates/storage/service.yaml
@@ -7,9 +7,13 @@ metadata:
   namespace: {{ .Values.ksNamespace }}
   annotations:
     {{- include "kubescape-operator.annotations" (dict "Values" .Values) | nindent 4 }}
+    {{- with .Values.storage.service.annotations }}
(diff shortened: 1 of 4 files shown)


matthyx (approved):
thanks @officialasishkumar

PR #825, opened 2026-04-26
ArgoCD GitOps Support for Kubescape Helm

lyuval-armosec:
## Summary

Adds opt-in GitOps/ArgoCD support to the `kubescape-operator` chart through two independent changes, plus minor template-layout refactors.

## Opt-in GitOps-friendly cert handling

A new top-level value selects how the admission-webhook and storage mTLS certificates are managed:

```yaml
certificates:
  strategy: template   # default — original behavior
  # strategy: hook     # GitOps-friendly
```

- **`template` (default)** — keeps the existing behavior: certs are generated at template-render time via `genCA` / `genSignedCert`, with a `lookup`-based fallback to preserve the CA across upgrades. No change for existing Helm users.
- **`hook`** — delegates cert lifecycle to two pre-install/pre-upgrade Helm hook Jobs per cert path (admission webhook + storage mTLS), backed by [`jkroepke/kube-webhook-certgen`](https://github.com/jkroepke/kube-webhook-certgen):
  - a `create` Job generates the CA + leaf cert into a Secret (idempotent — skips if the Secret already exists),
  - a `patch` Job stamps the resulting CA into the `ValidatingWebhookConfiguration` and `APIService` `caBundle`.

  Because these are standard Helm hooks (`pre-install`, `pre-upgrade`), they map cleanly onto ArgoCD's PreSync phase with no extra annotations or `ignoreDifferences` workarounds. Templates ship with empty `caBundle` and no Secret, so there's nothing for ArgoCD to drift against — the certs only exist at runtime.

This makes the chart fully compatible with ArgoCD (and other Helm-rendering GitOps controllers) when `hook` is selected, while preserving the simpler template-time path for plain `helm install` users.

Hooks are kept opt-in rather than made the default because they add a noticeable amount of time to every `helm install` / `helm upgrade` (mainly due to RBAC resources related to the hook jobs).

## Deterministic default cron schedules

Independently of the cert work, the default schedules for `kubescape-scheduler` and `kubevuln-scheduler` are now derived from a stable hash of `clusterName` / namespace / release name instead of `randNumeric`. Hour and minute are hashed with different prefixes to avoid correlation. The `lookup`-based "preserve existing schedule on upgrade" branch is kept for plain Helm.

This keeps the original goal (spread scan times across clusters to avoid a thundering herd) while giving every render the same output for a given installation — so ArgoCD no longer sees `/spec/schedule` drift on every reconcile. Users who set an explicit `scanSchedule` continue to bypass the helper entirely.

## Minor refactors

- Admission-webhook templates moved into their own `templates/operator/admission-webhook/` directory, with a dedicated `_helpers.tpl` (service domain, Secret/CA-Secret names, cert data builder).
- Storage cert helpers split out of `_helpersKubescapeStorage.tpl` into `templates/storage/certgen/_helpers.tpl` alongside the new hook resources.
- `imagePullSecrets` / `imagePullSecret` value docs clarified; required-fields validation added when generating the secret.
- `operator.admissionService` regrouped under `operator.admissionWebhooks.service` to make room for the new `patch` sub-tree.
- Snapshot tests extended to cover both strategies.

## Summary by CodeRabbit

* **New Features**
  * New certificates.strategy option (template vs hook) enabling runtime hook-based certificate generation and patching.

* **Improvements**
  * Admission webhook and storage mTLS flows updated for hook lifecycle with RBAC/jobs, deterministic daily scan scheduling, stricter image-pull-secret rendering, conditional pod annotations, and unified TLS secret naming.
  * ArgoCD Application values moved inline and certain diff-ignores removed; automated sync (prune:false, selfHeal:true) preserved.

* **Documentation**
  * Expanded GitOps guidance on certificate strategy.

* **Tests**
  * Added snapshot tests covering strategy and capability permutations.


charts/kubescape-operator/templates/operator/admission-webhook/admission-service.yaml
```diff
@@ -1,24 +1,22 @@
...
+    {{- with .Values.operator.admissionWebhooks.service.annotations }}
     {{- toYaml . | nindent 4 }}
     {{- end }}
   labels:
     {{- include "kubescape-operator.labels" (dict "Chart" .Chart "Release" .Release "Values" .Values "app" .Values.operator.name "tier" .Values.global.namespaceTier) | nindent 4 }}
 spec:
   ports:
-    - port: {{ .Values.operator.admissionService.port }}
-      targetPort: {{ .Values.operator.admissionService.targetPort }}
+    - port: {{ .Values.operator.admissionWebhooks.service.port }}
```
matthyx: #823 was additive here, but this PR removes `operator.admissionService.*` entirely. Any existing values files that override `type`, `port`, `targetPort`, or `annotations` under the old path will silently lose those overrides on upgrade. Can we coalesce the new keys with the legacy ones for a release or two and mark `operator.admissionService` deprecated in `values.yaml`/README?

charts/kubescape-operator/templates/operator/admission-webhook/_helpers.tpl
```diff
@@ -0,0 +1,48 @@
...
+{{- define "kubescape-admission.svcDomainName" -}}
+{{- printf "kubescape-admission-webhook.%s.svc" .Values.ksNamespace -}}
+{{- end }}
+
+{{- define "kubescape-admission.certificateData" -}}
+{{- $svcDomainName := (include "kubescape-admission.svcDomainName" .) -}}
+{{- $ca := dict "Key" "mock-ca-key" "Cert" "mock-ca-cert" -}}
+{{- $cert := dict "Key" "mock-cert-key" "Cert" "mock-cert-cert" -}}
+{{- if not .Values.unittest }}
+  {{- $existingCASecret := (lookup "v1" "Secret" .Values.ksNamespace (include "kubescape-admission.caSecretName" .)) -}}
```
matthyx: #823 explicitly fell back to the existing `kubescape-admission-ca` / `<svc>.svc-kubescape-tls-pair` secrets so upgrades could keep the old CA. This helper now only looks up the new `kubescape-admission-webhook-ca`, which means template-mode upgrades from current releases will mint a new CA/leaf pair and orphan the old secrets. That seems like a behavior change for existing Helm users unless we keep the legacy lookup/migration path here.

charts/kubescape-operator/templates/storage/certgen/job-patch.yaml
```diff
@@ -0,0 +1,87 @@
+{{- $components := fromYaml (include "components" .) }}
+{{- if and $components.storage.enabled .Values.storage.mtls.enabled (eq (include "kubescape.certificates.strategy" .) "hook") }}
+{{- $registry := .Values.storage.mtls.certgen.image.registry }}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name:  {{ include "storage.certgen.name" . }}-patch
+  namespace: {{ .Values.ksNamespace }}
+  annotations:
+    "helm.sh/hook": post-install,post-upgrade
```
matthyx, line 10: With `certificates.strategy=hook`, `templates/storage/apiservice.yaml` intentionally renders the APIService without `caBundle`. Under Argo CD, `post-install`/`post-upgrade` Helm hooks are treated as PostSync, and PostSync waits for the app to become healthy. An APIService without a valid CA stays Degraded, so this patch job may never run in the GitOps flow this PR is targeting. Can this patch stay in the pre-install/pre-upgrade lane (or get default ArgoCD PreSync annotations) so the CA is stamped before health gating kicks in?

charts/kubescape-operator/templates/operator/admission-webhook/clusterrole.yaml
```diff
@@ -0,0 +1,17 @@
+{{- $components := fromYaml (include "components" .) }}
+{{- if and $components.operator.enabled (eq .Values.capabilities.admissionController "enable") (eq (include "kubescape.certificates.strategy" .) "hook") }}
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: {{ include "kubescape-admission.name" . }}
```
matthyx: `metadata.name` is cluster-scoped here, but `kubescape-admission.name` is a fixed string. Two installs in different namespaces will race on the same ClusterRole/Binding and one uninstall/upgrade can break the other. Can we make the hook RBAC name release- or namespace-qualified here? The storage certgen ClusterRole/Binding have the same issue.

charts/kubescape-operator/templates/operator/admission-webhook/_helpers.tpl
lyuval-armosec: I can align the `template` version of the secret names to what it used to be, so existing users can reuse the old certs. The new names just seemed more consistent to me.

charts/kubescape-operator/templates/storage/certgen/job-patch.yaml
lyuval-armosec: From my tests running ArgoCD syncs the APIServer becomes `Synced` and the hooks move on. From what I observed it transitions to Healthy on fresh installs quite quickly.
We can't move the patch job to a pre-install/pre-upgrade since fresh installs won't have the APIService yet for it to update.

charts/kubescape-operator/templates/operator/admission-webhook/clusterrole.yaml
lyuval-armosec: I can make these namespace qualified

charts/kubescape-operator/templates/operator/admission-webhook/admission-service.yaml
lyuval-armosec: will do

charts/kubescape-operator/templates/operator/admission-webhook/clusterrole.yaml
lyuval-armosec: fixed

charts/kubescape-operator/templates/operator/admission-webhook/admission-service.yaml
lyuval-armosec: fixed

charts/kubescape-operator/templates/operator/admission-webhook/_helpers.tpl
lyuval-armosec: Regarding this point, from what i am aware helm doesn't orphan the last release's resources it cleans them up. This is the same for argocd (requires sync with prune to actually delete though).
Let me know if you still want me to support legacy naming

charts/kubescape-operator/templates/operator/admission-webhook/_helpers.tpl
```diff
@@ -0,0 +1,54 @@
...
+{{- printf "kubescape-admission-webhook.%s.svc" .Values.ksNamespace -}}
+{{- end }}
+
+{{- define "kubescape-admission.certificateData" -}}
+{{- if not .Values.global._admissionCertData -}}
+  {{- $svcDomainName := (include "kubescape-admission.svcDomainName" .) -}}
+  {{- $ca := dict "Key" "mock-ca-key" "Cert" "mock-ca-cert" -}}
+  {{- $cert := dict "Key" "mock-cert-key" "Cert" "mock-cert-cert" -}}
+  {{- if not .Values.unittest }}
+    {{- $existingCASecret := (lookup "v1" "Secret" .Values.ksNamespace (include "kubescape-admission.caSecretName" .)) -}}
```
matthyx, line 36: This `lookup` only checks the renamed CA secret, so the first upgrade from released versions skips the existing `kubescape-admission-ca` and mints a new CA.

Suggested edit: if `kubescape-admission-webhook-ca` is absent here, fall back to `lookup ... "kubescape-admission-ca"` before generating a new CA, then reuse that CA to sign the new leaf cert. That keeps the webhook trust chain stable across the first upgrade while still letting the chart converge on the new secret names afterward.

lyuval-armosec: will do

lyuval-armosec: @matthyx  added fallback

matthyx (approved):
looks good now, thanks @lyuval-armosec !

PR #841, opened 2026-05-13
fix: add liveness and readiness probes to grype-offline-db deployment

### LINKED ISSUE #840: fix(grype-offline-db): add liveness and readiness probes to deployment
## Description

The `grype-offline-db` deployment has no `livenessProbe` or `readinessProbe` defined, while every other deployment in the chart defines both probes.

| Deployment | livenessProbe | readinessProbe |
|---|---|---|
| kubescape | yes | yes |
| kubevuln | yes | yes |
| operator | yes | yes |
| otel-collector | yes | yes |
| prometheus-exporter | yes | yes |
| storage | yes | yes |
| grype-offline-db | no | no |

## Impact

`grype-offline-db` is an HTTP file server on port 8080 that serves the offline Grype vulnerability database. `kubevuln` fetches from it at scan time via `grypeDbListingURL`.

Without probes, if the HTTP server process crashes or hangs inside the container, Kubernetes sees the pod as Running and never restarts it. `kubevuln` then silently fails to fetch the database and produces incomplete vulnerability scan results with no visible indication to the user. Traffic also keeps routing to the dead server since it is never removed from Service endpoints.

## Expected Behavior

Add `livenessProbe` and `readinessProbe` using `tcpSocket` on port 8080, consistent with how other components define probes. Optionally expose these under `grypeOfflineDB` in `values.yaml` for user configurability.

## Verification

```bash
grep -c "readinessProbe\|livenessProbe" \
  charts/kubescape-operator/templates/grype-offline-db/deployment.yaml
# returns 0
```

Shreya2005-2005:
## Overview
The `grype-offline-db` deployment was missing both `livenessProbe` and `readinessProbe`, while every other deployment in the chart defines them. Since `grype-offline-db` runs an HTTP file server on port 8080 that `kubevuln` depends on for vulnerability database fetches, a crashed or hung server process goes undetected — Kubernetes keeps the pod as Running and kubevuln silently fails to fetch the DB, producing incomplete scan results.

Added `livenessProbe` and `readinessProbe` using `tcpSocket` on port 8080, consistent with the `prometheus-exporter` component which uses the same probe type.

## How to Test
```bash
helm template charts/kubescape-operator/ \
  --set clusterName=test \
  --set grypeOfflineDB.enabled=true \
  | grep -A10 "livenessProbe\|readinessProbe"
```

Snapshot tests:
```bash
docker run --rm -v "$(pwd)":/apps helmunittest/helm-unittest charts/kubescape-operator/
```
Expect 23/23 tests, 974/974 snapshots passing.

## Related issues/PRs
Closes #840

## Summary by CodeRabbit

* **Chores**
  * Improved deployment stability with health monitoring probes that enable faster detection and automatic recovery of container failures for enhanced service reliability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/helm-charts/pull/841)

### DIFF
--- a/charts/kubescape-operator/templates/grype-offline-db/deployment.yaml
+++ b/charts/kubescape-operator/templates/grype-offline-db/deployment.yaml
@@ -54,6 +54,16 @@ spec:
           ports:
           - containerPort: 8080
             protocol: TCP
+          livenessProbe:
+            tcpSocket:
+              port: 8080
+            initialDelaySeconds: 3
+            periodSeconds: 3
+          readinessProbe:
+            tcpSocket:
+              port: 8080
+            initialDelaySeconds: 3
+            periodSeconds: 3
           resources:
 {{ toYaml .Values.grypeOfflineDB.resources | indent 12 }}
       nodeSelector:
--- a/charts/kubescape-operator/templates/grype-offline-db/deployment.yaml
+++ b/charts/kubescape-operator/templates/grype-offline-db/deployment.yaml
@@ -55,12 +55,14 @@ spec:
           - containerPort: 8080
             protocol: TCP
           livenessProbe:
-            tcpSocket:
+            httpGet:
+              path: /databases
               port: 8080
             initialDelaySeconds: 3
             periodSeconds: 3
           readinessProbe:
-            tcpSocket:
+            httpGet:
+              path: /databases
               port: 8080
             initialDelaySeconds: 3
             periodSeconds: 3
--- a/charts/kubescape-operator/tests/__snapshot__/snapshot_test.yaml.snap
+++ b/charts/kubescape-operator/tests/__snapshot__/snapshot_test.yaml.snap
@@ -519,19 +519,21 @@ all capabilities:
             - image: quay.io/kubescape/grype-offline-db:latest
               imagePullPolicy: Always
               livenessProbe:
+                httpGet:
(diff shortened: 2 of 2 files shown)


charts/kubescape-operator/templates/grype-offline-db/deployment.yaml
```diff
@@ -54,6 +54,16 @@ spec:
           ports:
           - containerPort: 8080
             protocol: TCP
+          livenessProbe:
+            tcpSocket:
```
matthyx: Using a tcpSocket probe here only proves port 8080 accepted a TCP handshake. It does not prove the HTTP server kubevuln depends on is actually responding, so a hung or wedged HTTP worker can still stay Ready while DB fetches keep failing. Can we switch this (and the readiness probe below) to httpGet on a path the image serves instead of a raw TCP check?

Shreya2005-2005: Switched both `livenessProbe` and `readinessProbe` from `tcpSocket` to `httpGet` on `/databases` port 8080 — the same path kubevuln uses to fetch the DB listing, so a hung HTTP worker will now be detected. All 23 tests and 974 snapshots pass.

matthyx (approved):
thanks

PR #842, opened 2026-05-15
Fix kubevuln tmp-dir PVC rendering when disabled

aaa-aashna:
## Overview

Fix inconsistent kubevuln component gating for `tmp-dir-pvc.yaml`.

Previously, `charts/kubescape-operator/templates/kubevuln/tmp-dir-pvc.yaml` rendered a PersistentVolumeClaim based only on:

```yaml id="w4q8zp"
.Values.kubevuln.tmpDirPersistence.enabled
```

Unlike the other kubevuln templates in the same directory, it did not verify whether the kubevuln component itself was enabled through the shared `components` helper.

As a result, the PVC could still render even when kubevuln was disabled through capability configuration such as:

```yaml id="p7m2qx"
capabilities:
  vulnerabilityScan: disable
```

or:

```yaml id="u4v8wc"
capabilities:
  backend-storage: enable
```

This created inconsistent partial rendering behavior where:

* kubevuln Deployments/RBAC/resources were correctly skipped
* but the temporary PVC was still created and managed by the release

The fix aligns `tmp-dir-pvc.yaml` with the existing gating pattern already used by the other kubevuln templates by checking both:

* `kubevuln.enabled`
* `kubevuln.tmpDirPersistence.enabled`

This prevents orphan PVC creation when kubevuln is logically disabled.

## How to Validate

Example repro values:

```yaml id="k5n1vr"
capabilities:
  vulnerabilityScan: disable

kubevuln:
  tmpDirPersistence:
    enabled: true
```

Before this change:

* `tmp-dir-pvc.yaml` still rendered a PVC

After this change:

* the PVC is no longer rendered when kubevuln is disabled

## Checklist

* [x] My code follows the style guidelines of this project
* [x] I have performed a self-review of my code
* [x] The change is localized and backward compatible

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced PersistentVolumeClaim deployment logic to ensure temporary directory storage is only provisioned when both the persistence feature and the kubevuln component are properly enabled, improving deployment consistency.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/helm-charts/pull/842)

### DIFF
--- a/charts/kubescape-operator/templates/kubevuln/tmp-dir-pvc.yaml
+++ b/charts/kubescape-operator/templates/kubevuln/tmp-dir-pvc.yaml
@@ -1,4 +1,5 @@
-{{- if .Values.kubevuln.tmpDirPersistence.enabled }}
+{{- $components := fromYaml (include "components" .) }}
+{{- if and $components.kubevuln.enabled .Values.kubevuln.tmpDirPersistence.enabled }}
 apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:


aaa-aashna:
@matthyx any update about this one ?

matthyx (approved):

PR #843, opened 2026-05-17
certificate strategy hook -> initContainer

lyuval-armosec:
**Title:** `fix: replace Helm hook cert generation with init-container strategy`

---

**Summary**

Replaces the `kube-webhook-certgen` Helm hook Jobs for the admission webhook (`ValidatingWebhookConfiguration`) and storage mTLS (`APIService`) with a pair of init containers running on the consuming pod. This avoids the GitOps/ArgoCD deadlock where Helm hook Jobs do not run until the release is healthy, but the release cannot become healthy until the `caBundle` fields are patched.

The new approach reuses the existing `grypeOfflineDB.rollout.image` (`quay.io/kubescape/kubectl`) and inlines the `certgen-create.sh` / `certgen-patch.sh` logic in the chart templates, so no separate `kube-init-certgen` image or operator-side release pipeline is required.

**Changes**

- `certificates.strategy` valid values changed from `[template, hook]` to `[template, initContainer]`
- Deleted `job-create.yaml`, `job-patch.yaml`, `serviceaccount.yaml` for both admission-webhook and storage/certgen
- Added `initContainers` blocks to the operator and storage Deployments, gated on `strategy=initContainer`
- `tls-certs` / `ca-certificates` volumes switch to `emptyDir` under `initContainer` strategy; the Secret is managed by the init container, not Helm
- RBAC (Role, ClusterRole, RoleBinding, ClusterRoleBinding) is rebound from the deleted ServiceAccounts to the existing component ServiceAccounts
- `operator.admissionWebhooks.patch` and `storage.mtls.certgen` now keep `retries` / `retryIntervalSeconds`; the init-container image is reused from `grypeOfflineDB.rollout.image`

**Migration note**

Users on `certificates.strategy: hook` must update to `certificates.strategy: initContainer`. Existing `*-tls-hook` Secrets will be orphaned; the init container creates new `*-tls-init` Secrets on first run.

## Summary by CodeRabbit

* **New Features**
  * Added initContainer-based certificate generation flow with runtime scripts and ConfigMaps to create and patch TLS material.
* **Chores**
  * Replaced hook-based jobs/serviceaccounts with initContainer behavior; updated secret naming, pod checksum annotations, and values configuration for certgen.
* **Bug Fixes**
  * Adjusted RBAC verbs for webhook and apiservice resources to narrow permissions.
* **Tests**
  * Updated snapshot tests to exercise template vs initContainer certificate strategies.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/helm-charts/pull/843?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


charts/kubescape-operator/templates/operator/admission-webhook/configmap.yaml
```diff
@@ -1,14 +1,17 @@
 {{- $components := fromYaml (include "components" .) }}
-{{- if and $components.operator.enabled (eq .Values.capabilities.admissionController "enable") (eq (include "kubescape.certificates.strategy" .) "hook") }}
-kind: ServiceAccount
+{{- if and $components.operator.enabled (eq .Values.capabilities.admissionController "enable") (eq (include "kubescape.certificates.strategy" .) "initContainer") }}
 apiVersion: v1
+kind: ConfigMap
```
matthyx, line 4: These ConfigMaps are now carrying executable content. Can we mark them `immutable: true` here (and in `templates/storage/certgen/configmap.yaml`) so a namespace-level `patch/update` on ConfigMaps cannot silently swap the script payload that will run on the next pod restart?

charts/kubescape-operator/templates/operator/deployment.yaml
```diff
@@ -59,6 +59,73 @@ spec:
...
+            - /etc/certs
+            - --ca-name
+            - ca.crt
+            - --cert-name
+            - tls.crt
+            - --key-name
+            - tls.key
+          volumeMounts:
+            - name: admission-certgen-scripts
+              mountPath: /scripts
```
matthyx, line 99: Can we make the `/scripts` mount explicit `readOnly: true` here (same for the second init container and the storage deployment)? Projected ConfigMap volumes are typically read-only anyway, but the explicit flag is good defense-in-depth and makes the intent clear to policy engines.

charts/kubescape-operator/templates/operator/deployment.yaml
```diff
@@ -202,10 +269,19 @@ spec:
             secretName: {{ .Values.global.extraCaCertificates.secretName }}
         {{- end }}
         {{- if eq .Values.capabilities.admissionController "enable" }}
+        {{- if eq (include "kubescape.certificates.strategy" .) "initContainer" }}
+        - name: tls-certs
+          emptyDir: {}
+        - name: admission-certgen-scripts
+          configMap:
```
matthyx, line 281: Since the pod now executes code out of this ConfigMap, can we add a `checksum/*` pod-template annotation for the script content (and mirror it in `storage/deployment.yaml`)? Without that, script changes do not trigger a rollout and the drift stays silent until an unrelated restart.

matthyx (commented):
Left a few configmap hardening notes inline: immutable script ConfigMaps, explicit readOnly mounts, and a rollout checksum for script changes.

charts/kubescape-operator/scripts/certgen-create.sh
```diff
@@ -0,0 +1,105 @@
...
+mkdir -p "$OUT_DIR"
+
+# --- path 1: secret already exists -> just materialize cert / key ---
+if SECRET_JSON="$(kubectl -n "$NAMESPACE" get secret "$SECRET_NAME" -o json 2>/dev/null)"; then
+  echo "secret $NAMESPACE/$SECRET_NAME exists, reusing"
+  echo "$SECRET_JSON" | jq -er --arg k "$CERT_NAME" '.data[$k] // empty' | base64 -d > "$OUT_DIR/$CERT_NAME"
+  echo "$SECRET_JSON" | jq -er --arg k "$KEY_NAME"  '.data[$k] // empty' | base64 -d > "$OUT_DIR/$KEY_NAME"
+  [[ -s "$OUT_DIR/$CERT_NAME" && -s "$OUT_DIR/$KEY_NAME" ]] || { echo "secret missing $CERT_NAME/$KEY_NAME" >&2; exit 1; }
+  exit 0
+fi
```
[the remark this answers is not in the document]
lyuval-armosec: singleton deployments

charts/kubescape-operator/templates/operator/admission-webhook/configmap.yaml
lyuval-armosec: will do

charts/kubescape-operator/templates/operator/deployment.yaml
lyuval-armosec: yes sir

charts/kubescape-operator/templates/operator/deployment.yaml
lyuval-armosec: sure thing

matthyx (changes requested):
Thanks for pushing this through. The init-container approach makes sense, but I think there are three blocking upgrade and deployability regressions:

1. Existing `certificates.strategy=hook` installs can no longer upgrade. `_common.tpl` now rejects `hook`, and `helm template ... --set certificates.strategy=hook` fails. That breaks current hook users, including the `ArgoCDApplication.yaml` example in this repo.

2. The new `*-certgen-scripts` ConfigMaps are `immutable: true` with fixed names. The first future script change will make `helm upgrade` fail because Helm will try to patch immutable ConfigMaps in place.

3. Existing certgen image override keys were removed (`operator.admissionWebhooks.patch.image.*` and `storage.mtls.certgen.image.*`). Those old values are now ignored and the init containers always render `quay.io/kubescape/kubectl:1.36.1`, which breaks private-registry and mirror setups on upgrade.

Please preserve backward compatibility or a migration path for hook users and the old image override keys, and either drop `immutable: true` or content-version the ConfigMap names.

matthyx (dismissed):

lyuval-armosec:
> Thanks for pushing this through. The init-container approach makes sense, but I think there are three blocking upgrade and deployability regressions:
> 
> 1. Existing `certificates.strategy=hook` installs can no longer upgrade. `_common.tpl` now rejects `hook`, and `helm template ... --set certificates.strategy=hook` fails. That breaks current hook users, including the `ArgoCDApplication.yaml` example in this repo.
> 2. The new `*-certgen-scripts` ConfigMaps are `immutable: true` with fixed names. The first future script change will make `helm upgrade` fail because Helm will try to patch immutable ConfigMaps in place.
> 3. Existing certgen image override keys were removed (`operator.admissionWebhooks.patch.image.*` and `storage.mtls.certgen.image.*`). Those old values are now ignored and the init containers always render `quay.io/kubescape/kubectl:1.36.1`, which breaks private-registry and mirror setups on upgrade.
> 
> Please preserve backward compatibility or a migration path for hook users and the old image override keys, and either drop `immutable: true` or content-version the ConfigMap names.

I updated README & argocd example, with mention for hook users to use `initContainer` cert startegy.
I also kept configmaps as immutable and created content-versioned names for the configmaps.

matthyx (approved):

Issue #851, opened 2026-06-01
`node-agent` configmap silently disables `nodeProfileServiceEnabled` and `networkStreamingEnabled` on community installs, the chart's `capabilities` block says they're on, the rendered config says they're off


yugal07:
# Description

The chart's `capabilities` block is presented to users as the single source of truth for which node-agent features are on. The rendered `node-agent` configmap, however, ANDs each capability with a *second*, internal precondition that is not visible from `values.yaml`:

```yaml
# helm-charts/charts/kubescape-operator/templates/node-agent/configmap.yaml:34-35
"nodeProfileServiceEnabled": {{ and $components.synchronizer.enabled (eq .Values.capabilities.nodeProfileService "enable") }},
"networkStreamingEnabled":   {{ and $configurations.submit           (eq .Values.capabilities.networkEventsStreaming "enable") }},
```

So:

- `capabilities.nodeProfileService=enable` only takes effect if **synchronizer is enabled**. On the default community install (no ARMO account) the synchronizer is disabled, and the capability collapses to `false` without any warning.
- `capabilities.networkEventsStreaming=enable` only takes effect if **configurations.submit is true**. Same story: community installs (no `account` set → `submit=false`) get the capability silently dropped.

Observed on a fresh chart 1.40.2 install with default values + `clusterName=…`:

```
$ kubectl -n kubescape get cm ks-capabilities -o json | jq -r '.data.capabilities' | jq '.capabilities | {nodeProfileService, networkEventsStreaming}'
{
  "nodeProfileService": "enable",
  "networkEventsStreaming": "enable"
}

$ kubectl -n kubescape get cm node-agent -o json | jq -r '.data["config.json"]' | jq '{nodeProfileServiceEnabled, networkStreamingEnabled}'
{
  "nodeProfileServiceEnabled": false,
  "networkStreamingEnabled":   false
}
```

Two configmaps in the same namespace, generated by the same chart, disagree on whether the feature is enabled. The user has no way to know:

- They didn't set anything to `disable`.
- `helm get values` still shows `nodeProfileService: enable`.
- The chart's NOTES output and the per-component READMEs reference the capability flag and assume "enable" is the truth.
- The first place the truth lives is the `node-agent` configmap, mounted at `/etc/config/config.json` inside the DaemonSet — not somewhere a normal user thinks to look.

[...]

PR #852, opened 2026-06-03
warn when capability flags are dropped without a backend

yugal07:
NodeProfileService and networkEventsStreaming default to "enable" and are reported as such in the ks-capabilities configmap, but the node-agent configmap silently ANDs them with the synchronizer/submit precondition and renders them false on backend-less (community) installs. The two configmaps disagree with no warning to the user.

Add NOTES.txt warnings (mirroring the existing nodeSbomGeneration warning) that fire when either capability is "enable" but submit is disabled, pointing users at .Values.server / account / accessKey.

#851 - Approach as discussed  with @matthyx on slack, Could also go with the other results as discussed in the issue ticket expected behaviours if that is preferred. Upto the code reviewers. All 3 approaches mentioned are doable. 

  ### Testing
  - **Community install** (no `server`): both warnings render in NOTES. ✅
  - **Backend configured** (`server`+`account`+`accessKey`): no warnings, and the
    node-agent flags render `true` — warning fires exactly when the flag would be `false`. ✅
  - `helm lint` passes; chart renders cleanly in both configurations. ✅

## Summary by CodeRabbit

* **New Features**
  * Config now exposes computed effective capability flags and capability warnings so installations and agents reflect actual enabled features.

* **Documentation**
  * Installation notes now surface warnings when optional features (node profiling, network event streaming, etc.) cannot be enabled due to missing backend or prerequisites, with guidance for remediation.

### DIFF
--- a/charts/kubescape-operator/templates/NOTES.txt
+++ b/charts/kubescape-operator/templates/NOTES.txt
@@ -49,7 +49,7 @@ To enable scanning of private images, either:
 WARNING: capabilities.nodeProfileService is set to "enable", but the backend is not configured (.Values.server is empty / submit is disabled).
 This capability requires the synchronizer, which is only enabled when connected to a backend.
 As a result, nodeProfileServiceEnabled will render as FALSE in the node-agent configmap and node profiles will NOT be generated.
-To use this capability, connect to a backend by setting `.Values.server` (and `account` / `accessKey`).
+To use this capability, connect to a backend by setting `.Values.server`. This requires `account` and `accessKey` (or an existing `credentials.cloudSecret`); when submit is enabled and no cloudSecret is provided, the chart will fail if these are missing.
 {{- end }}
 
 {{- if and (eq .Values.capabilities.networkEventsStreaming "enable") (not $configurations.submit) }}
@@ -57,7 +57,7 @@ To use this capability, connect to a backend by setting `.Values.server` (and `a
 WARNING: capabilities.networkEventsStreaming is set to "enable", but the backend is not configured (.Values.server is empty / submit is disabled).
 This capability requires submitting data to a backend.
 As a result, networkStreamingEnabled will render as FALSE in the node-agent configmap and network events will NOT be streamed.
-To use this capability, connect to a backend by setting `.Values.server` (and `account` / `accessKey`).
+To use this capability, connect to a backend by setting `.Values.server`. This requires `account` and `accessKey` (or an existing `credentials.cloudSecret`); when submit is enabled and no cloudSecret is provided, the chart will fail if these are missing.
 {{- end }}
 
 {{- $legacyAdmissionSvc := .Values.operator.admissionService | default dict }}
--- a/charts/kubescape-operator/templates/NOTES.txt
+++ b/charts/kubescape-operator/templates/NOTES.txt
@@ -1,5 +1,6 @@
 Thank you for installing {{ .Chart.Name }} version {{ .Chart.Version }}.
 {{ $components := fromYaml (include "components" .) -}}
+{{ $configurations := fromYaml (include "configurations" .) -}}
 {{ if $components.kubescapeScheduler.enabled -}}
 
 View your cluster's configuration scanning schedule:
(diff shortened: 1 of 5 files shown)


yugal07:
tests are failing, let me regenerate the snapshot

matthyx:
Reviewed locally against the PR head (built deps, `helm unittest` → 966 snapshots pass, `helm lint` clean). The NOTES warnings render exactly in the backend-less default case (`minimal capabilities` snapshot) and are absent in every server-configured case, which matches the intent. **No blockers** — the change is correct as written. A few non-blocking notes:

**1. The `nodeProfileService` warning keys off a different expression than the gate it describes (worth fixing).**
The node-agent configmap gates this flag on the synchronizer:
```
"nodeProfileServiceEnabled": {{ and $components.synchronizer.enabled (eq .Values.capabilities.nodeProfileService "enable") }}
```
but the new warning fires on `(not $configurations.submit)`. These are equivalent *today* only because `_common.tpl` defines `synchronizer.enabled: {{ $configurations.submit }}`. If that gate ever changes (e.g. synchronizer becomes independently toggleable), the warning silently drifts out of sync with the rendered value — which is exactly the bug this PR is fixing. Suggest gating the `nodeProfileService` warning on the same `$components.synchronizer.enabled` expression the configmap uses, so the two can't diverge. (`networkEventsStreaming` already matches its gate — both use `$configurations.submit` — so it's fine.)

**2. NOTES.txt isn't seen by GitOps installs.** The linked issue specifically calls out ArgoCD/Flux users reconciling values from git. Those flows render with `helm template`, which does not emit NOTES.txt, so they won't see these warnings. This is inherent to the "minimum" (NOTES) option from the issue and a fine starting point given the Slack discussion — just flagging that the GitOps use-case the issue highlighted isn't covered by this approach. A render-time signal (a `fail` for the contradictory case, or surfacing the effective value in a rendered manifest/configmap) would be needed to reach those users.

**3. Other flags with the same silent-drift shape are out of scope here.** The issue asked to audit every `{{ and <precondition> (eq .Values.capabilities.<x> "enable") }}` line. Besides the two covered, the same pattern drops `nodeScan` / `configurationScan` / `vulnerabilityScan` when `backend-storage` is enabled (`_common.tpl`) and `httpDetection` when `runtimeDetection` is disabled (`node-agent/configmap.yaml:27`), all without a warning. Reasonable to defer, but noting it so the audit isn't lost.

yugal07:
On it. Thanks

yugal07:
Pivoted this from the NOTES-only approach so it covers all three review points at once. The gate logic now lives in a single helper (capabilities.gates) that the node-agent configmap reads for its effective values, so the rendered flag and the warning can't drift apart anymore. The same helper feeds effectiveCapabilities and capabilityWarnings into the ks-capabilities configmap, which means GitOps users who reconcile with helm template (and never see NOTES) finally get the signal too.

I also extended the audit to httpDetection and reflected the backend-storage scans as backend instead of warning on them, since that one is intentional redirection. The node-agent config.json renders byte-identical to main, lint is clean, and all 966 unittest snapshots pass.

One thing worth your call: because the chart ships httpDetection: enable with runtimeDetection: disable, the httpDetection warning now fires on a default community install, so if you consider that noise I can drop it from the warned set or we can fix the default values separately.

matthyx:
Re-reviewed after the rework — this is a big improvement and addresses all three points from my earlier review:

- **No more drift.** The new `capabilities.gates` helper is a single source of truth: `node-agent/configmap.yaml` consumes `effective.*`, and both `ks-capabilities` and NOTES.txt consume `effectiveCapabilities`/`warnings` from the same helper. The `nodeProfileService` gate now keys off `synchronizer.enabled` (the actual gate), so the rendered value and the warning can no longer diverge — verified in the snapshots: `httpDetectionEnabled: false` co-occurs with both the matching `capabilityWarnings` entry and `effectiveCapabilities.httpDetection: disable`.
- **GitOps users are now covered.** Surfacing `effectiveCapabilities` + `capabilityWarnings` in the rendered `ks-capabilities` configmap means ArgoCD/Flux (`helm template`) installs see the drift too, not just interactive `helm install`. This was my main concern with the NOTES-only approach.
- **Broader audit included.** `httpDetection` (gated on `runtimeDetection`) now warns, and the backend-storage scans render as `backend` in `effectiveCapabilities`.

Verified locally on the PR head: `helm unittest` passes (966 snapshots), `helm lint` clean, CodeRabbit green.

**One blocker before merge: DCO is failing.** The last commit (`b14955b` "Coderabbit change") has no `Signed-off-by` line — it looks like it was applied through the GitHub UI, which doesn't add the trailer. Sign it off to make the check pass, e.g.:

```
git rebase --signoff main
git push --force-with-lease
```

(or amend just that commit with `git commit --amend -s`). Once DCO is green this LGTM from my side.

matthyx (approved):

PR #854, opened 2026-06-08
feat: add cveMatchingMode and trustedVendors to kubevuln config

slashben:
## What

Plumbs kubevuln's new adaptive CVE-matching settings through the kubescape-operator chart into kubevuln's `clusterData.json`. Two new knobs under `kubevuln.config`:

| Key | Default | Meaning |
|-----|---------|---------|
| `cveMatchingMode` | `""` (→ `adaptive`) | `off` (Grype defaults) / `on` (CPE matching everywhere) / `adaptive` (CPE on, except trusted-vendor images) |
| `trustedVendors` | `[]` | distro identifiers trusted in adaptive mode; empty uses kubevuln's built-in set (echo, chainguard, wolfi, minimos) |

Depends on the kubevuln change that adds `cveMatchingMode` (kubevuln PR #372). Adaptive only takes effect once an image with that support is pinned.

## Backward compatibility (the design goal)

The configmap resolves the effective mode and emits **both** the new `cveMatchingMode` key and a **derived** `useDefaultMatchers`, so an **old kubevuln image keeps working unchanged** — it ignores the unknown `cveMatchingMode`/`trustedVendors` keys (mapstructure drops unknown fields) and reads the derived boolean:

| values | `cveMatchingMode` | `useDefaultMatchers` | new image | old image |
|---|---|---|---|---|
| defaults | `adaptive` | `false` | adaptive | CPE on (unchanged) |
| `useDefaultMatchers: true` (legacy) | `off` | `true` | off | off (intent preserved) |
| `cveMatchingMode: on` | `on` | `false` | on | CPE on |
| `cveMatchingMode: off` | `off` | `true` | off | off |
| `cveMatchingMode: on` + `useDefaultMatchers: true` | `on` | `false` | on (explicit wins) | CPE on |

A legacy `useDefaultMatchers: true` is preserved as mode `off`; `adaptive` maps to `useDefaultMatchers: false` (CPE on) as the safe fallback for old binaries. An invalid `cveMatchingMode` fails `helm template` (with a clear message) rather than crashing kubevuln at boot.

## Deliberately out of scope
- **kubevuln image tag** — not bumped; adaptive activates when the supporting release is pinned (follow-up).
- **Chart version/appVersion** — left to the maintainers' release-prep process.

## Testing
- `helm lint` clean.
- `helm template` matrix verified (default, legacy bool, on, off, adaptive+trustedVendors, explicit-wins, invalid-guard); default `clusterData` parses as valid JSON; `trustedVendors` renders as a proper JSON array.
- `helm unittest charts/kubescape-operator/` passes (snapshot regenerated, 966 snapshots).
- README values table updated for the two new keys.

## Summary by CodeRabbit

* **New Features**
  * Added configurable CVE matching modes: off, on, and adaptive.
  * Added support for specifying trusted vendors in configuration.

* **Documentation**
  * Updated configuration documentation with new options and backward-compatibility details for existing settings.

### DIFF
--- a/charts/kubescape-operator/templates/configs/cloudapi-configmap.yaml
+++ b/charts/kubescape-operator/templates/configs/cloudapi-configmap.yaml
@@ -1,5 +1,15 @@
 {{- $components := fromYaml (include "components" .)  }}
 {{- $configurations := fromYaml (include "configurations" .)  }}
+{{- $cveMatchingModeRaw := .Values.kubevuln.config.cveMatchingMode }}
+{{- $cveMatchingMode := "" }}
+{{- if kindIs "bool" $cveMatchingModeRaw }}
+  {{- /* Unquoted off/on/yes/no/true/false in YAML are parsed as booleans (the "Norway problem"); normalize back to the intended mode instead of silently misreading it. */}}
+  {{- $cveMatchingMode = ternary "on" "off" $cveMatchingModeRaw }}
+{{- else if $cveMatchingModeRaw }}
+  {{- $cveMatchingMode = $cveMatchingModeRaw }}
+{{- end }}
+{{- if not $cveMatchingMode }}{{- $cveMatchingMode = ternary "off" "adaptive" .Values.kubevuln.config.useDefaultMatchers }}{{- end }}
+{{- if not (has $cveMatchingMode (list "off" "on" "adaptive")) }}{{- fail (printf "kubevuln.config.cveMatchingMode must be one of off|on|adaptive, got %q" $cveMatchingMode) }}{{- end }}
 kind: ConfigMap
 apiVersion: v1
 metadata:
@@ -31,7 +41,11 @@ data:
       "scanTimeout": "{{ .Values.kubevuln.config.scanTimeout }}",
       "scanEmbeddedSBOMs": {{ eq .Values.capabilities.scanEmbeddedSBOMs "enable" }},
       "vexGeneration": {{ eq .Values.capabilities.vexGeneration "enable" }},
-      "useDefaultMatchers": {{ .Values.kubevuln.config.useDefaultMatchers }},
+      "cveMatchingMode": "{{ $cveMatchingMode }}",
+      "useDefaultMatchers": {{ eq $cveMatchingMode "off" }},
+{{- if .Values.kubevuln.config.trustedVendors }}
+      "trustedVendors": {{ .Values.kubevuln.config.trustedVendors | toJson }},
+{{- end }}
       "storeFilteredSbom": {{ or .Values.kubevuln.config.storeFilteredSboms (eq .Values.capabilities.syncSBOM "enable") }},
       "continuousPostureScan": {{ $configurations.continuousScan }},
 {{- if not (empty .Values.kubevuln.config.grypeDbListingURL) }}
--- a/charts/kubescape-operator/README.md
+++ b/charts/kubescape-operator/README.md
@@ -146,6 +146,8 @@ However, we recommend that you give Kubescape no less than 500m CPU no matter th
 | kubevuln.volumeMounts | object | `[]` | Additional volumeMounts for the image vulnerability scanning |
 | kubevuln.config.grypeDbListingURL | string | `""` | Parameter to override the default Grype vulnerability database URL (listings.json format) |
 | kubevuln.config.proxyRegistryMap | object | `{}` | Map of source registry address to mirror registry address. When set, kubevuln rewrites image pull references accordingly (useful for air-gapped or offline registry mirrors) |
+| kubevuln.config.cveMatchingMode | string | `""` | CVE matching mode: "off" (Grype defaults), "on" (CPE matching everywhere), or "adaptive" (CPE matching everywhere except trusted-vendor images, which rely on the vendor's authoritative feed). Empty derives the mode from useDefaultMatchers for backward compatibility (true -> "off", false -> "adaptive"). |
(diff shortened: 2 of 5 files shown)


charts/kubescape-operator/templates/configs/cloudapi-configmap.yaml
```diff
@@ -1,5 +1,8 @@
 {{- $components := fromYaml (include "components" .)  }}
 {{- $configurations := fromYaml (include "configurations" .)  }}
+{{- $cveMatchingMode := .Values.kubevuln.config.cveMatchingMode | default "" }}
+{{- if not $cveMatchingMode }}{{- $cveMatchingMode = ternary "off" "adaptive" .Values.kubevuln.config.useDefaultMatchers }}{{- end }}
```
matthyx, line 11: Blocker: `.Values.kubevuln.config.cveMatchingMode` here can arrive as a **bool**, not a string, when a user writes the intuitive unquoted form. `cveMatchingMode: off` → YAML bool `false` → `default ""` treats it as empty → this `ternary` derives `adaptive` (the opposite of the user's intent). `cveMatchingMode: on` → bool `true` → passes through and trips the `has` check on the next line with `got %!q(bool=true)`. Consider `| toString` and/or normalizing a bool value to `on`/`off` so the unquoted form is safe. Verified on this branch with `helm template`.

matthyx (changes requested):
Thanks — the design is sound and the derived `useDefaultMatchers` backward-compat mapping renders correctly for every quoted scenario I tested (default→adaptive, legacy `useDefaultMatchers: true`→off, `"on"`, `"off"`, `"adaptive"`+trustedVendors, explicit-wins, invalid→clean fail; `helm lint` clean, output is valid JSON). One blocker before merge.

## Blocker: unquoted `cveMatchingMode: off` / `on` is coerced to a YAML boolean (Norway problem)

`off`, `on`, `yes`, `no`, `true`, `false` are parsed as booleans by Helm's YAML loader (YAML 1.1) unless quoted. The most intuitive way for a user to disable CPE matching — `cveMatchingMode: off` in a values file — silently does the **opposite** of what they asked, and `cveMatchingMode: on` hard-fails the whole render with a cryptic message. Verified against this branch (`helm template`, chart-built):

```yaml
# kubevuln.config.cveMatchingMode: off   (unquoted)
"cveMatchingMode": "adaptive",     # <-- user asked for off, got adaptive
"useDefaultMatchers": false,
```
```
# kubevuln.config.cveMatchingMode: on    (unquoted)
Error: ... kubevuln.config.cveMatchingMode must be one of off|on|adaptive, got %!q(bool=true)
```
(`"off"` / `"on"` quoted both work correctly.)

Why it happens: unquoted `off` → bool `false` → Sprig `default ""` treats `false` as empty → falls through to the `useDefaultMatchers`-derived branch → `adaptive`. Unquoted `on` → bool `true` → not empty, fails the `has` allow-list check, and the error renders the bool as `%!q(bool=true)` which doesn't hint at the real cause.

Suggested fixes (any one):
- Coerce before validating, e.g. `{{- $cveMatchingMode := .Values.kubevuln.config.cveMatchingMode | toString | default "" }}` and normalize a bool-typed value (`true`→`on`, `false`→`off`) so the intuitive form does the right thing; or
- Detect a non-string with `kindIs` and `fail` with a message that explicitly says "quote the value (`\"off\"`), `off`/`on` are parsed as booleans in YAML"; or
- At minimum, document in values.yaml/README that the value **must be quoted**.

Given this is security-scanning behavior, the silent `off`→`adaptive` case is the dangerous one and is worth guarding in-template rather than only in docs.

## Minor (non-blocking)
- The committed snapshot only exercises `off` (all-capabilities) and the derived `adaptive` (defaults). The matrix you list in the description — explicit `on`, `trustedVendors` array rendering, and the invalid-mode `fail` guard — isn't covered by a committed `helm unittest`. Worth adding so the guard/coercion can't silently regress.
- values.yaml line 456 comment says `false -> "on"` while the new default path derives `false -> adaptive`; the "for old kubevuln images" scoping is correct but a reader comparing the two comments may find it confusing — consider aligning the wording.

matthyx (approved):
Re-reviewed after the force-push (`923fa83`). The blocker is resolved. ✅

The new `kindIs "bool"` branch normalizes the YAML-coerced boolean back to a mode (`true`/`yes`→`on`, `false`/`no`→`off`), so the intuitive unquoted form now does the right thing. Verified against this commit with `helm template --kube-version 1.30`:

| values | cveMatchingMode | useDefaultMatchers |
|---|---|---|
| `cveMatchingMode: off` (unquoted) | `off` | `true` ✅ (was silently `adaptive`) |
| `cveMatchingMode: on` (unquoted) | `on` | `false` ✅ (was a hard render failure) |
| `cveMatchingMode: yes` / `no` | `on` / `off` | `false` / `true` ✅ |
| `"off"` quoted | `off` | `true` ✅ |
| default | `adaptive` | `false` ✅ |
| legacy `useDefaultMatchers: true` | `off` | `true` ✅ |
| `adaptive` + `trustedVendors` | `adaptive` | `false`, array rendered ✅ |
| `cveMatchingMode: enabled` | — | clean `fail` ✅ |

`helm lint` clean, `helm unittest` passes (37 tests, 966 snapshots), default `clusterData` is valid JSON, and the values.yaml `useDefaultMatchers` comment is now consistent (`false -> "adaptive"`). LGTM.

Non-blocking nit (optional): the committed snapshot still only exercises `off` and the derived `adaptive`; a small `helm unittest` asserting `on`, the `trustedVendors` array, and the invalid-mode `fail` would lock in the new normalization/guard against future regressions. Not required for merge.

PR #855, opened 2026-06-11
feat(operator): add opt-in capabilities.remediation with gated mutating RBAC

yugal07:
# helm-charts: add opt-in `capabilities.remediation` + gated operator RBAC

**Related issue:** [kubescape/kubescape#1770 — Kubescape CLI control over cluster operations](https://github.com/kubescape/kubescape/issues/1770)
**Design:** [kubescape/designs-and-proposals#5 — Kubescape CLI Control over Cluster Operations](https://github.com/kubescape/designs-and-proposals/pull/5) (merged)
**Pairs with:** https://github.com/kubescape/operator/pull/383

## What this PR does

The operator's Phase-1 remediation needs mutating RBAC the default install does
not grant (the operator ClusterRole is read-only on workloads/pods/nodes and has
no `networkpolicies`). This PR adds that RBAC **behind a new opt-in capability
flag, default off**, so the default install gains no new powers.

- **`values.yaml`** — new `capabilities.remediation: disable`, documented as an
  opt-in capability with elevated security risk that grants the operator mutating
  RBAC.
- **`templates/operator/clusterrole.yaml`** — a rule block gated on
  `{{- if eq .Values.capabilities.remediation "enable" }}`:
  - `apps` deployments/statefulsets/daemonsets → `patch` (backs `annotate`),
  - core pods/nodes → `patch` (`annotate`, and `cordon` in a later phase),
  - `networking.k8s.io` networkpolicies → `create`/`delete`/`get`/`list`
    (`quarantine`, later phase).
- **`tests/snapshot_test.yaml`** — set `remediation: enable` in the three
  "everything-on" scenarios, plus **two focused tests** asserting the RBAC is
  present when enabled and absent in the default install.
- **`tests/__snapshot__/snapshot_test.yaml.snap`** — regenerated: the
  `components-configmap` (the operator's runtime config) now advertises the
  `remediation` capability.

## Why gated and default-off

This is the design's headline safety property: "all destructive capability is
opt-in at install time." RBAC is the *hard* enforcement gate — with
`remediation: disable`, the operator literally cannot patch workloads, so even a
mistakenly-triggered action fails closed with `Forbidden` (recorded on the
`OperatorCommand` status), mutating nothing.

The RBAC block provisions for the whole Phase 1–3 action set in one place so the
later quarantine/cordon PRs do not need to reopen the ClusterRole; every verb is
commented with the action it serves.

## Cross-component compatibility (verified)

Adding `remediation` to `values.capabilities` makes it appear in the
`components-configmap` that the operator reads at startup. Verified that the
operator's real `LoadCapabilitiesConfig` parses the chart-rendered
`capabilities.json` (with `remediation`) without error and still populates the
known fields — unknown keys are ignored (viper default), exactly as the existing
`riskAcceptance` key already is. The operator does not need a new config field for
Phase 1: RBAC is the gate.

## Testing

- `helm lint` — clean (only the cosmetic "icon recommended" INFO).
- `helm unittest` — **28 tests / 966 snapshots pass** after regeneration.
- Rendered the operator ClusterRole both ways: read-only by default; with
  `capabilities.remediation=enable`, exactly the three mutating rules appear.

## Phasing

Phase 1 RBAC. The same flag/block already covers the verbs that Phase 2
(`quarantine`, networkpolicies) and Phase 3 (`cordon`, node patch) will exercise,
so those PRs add operator behaviour only.

### DIFF
--- a/charts/kubescape-operator/templates/operator/clusterrole.yaml
+++ b/charts/kubescape-operator/templates/operator/clusterrole.yaml
@@ -49,8 +49,12 @@ rules:
   {{- if eq .Values.capabilities.remediation "enable" }}
   # Post-scan remediation (opt-in). These mutating verbs are granted only when
   # capabilities.remediation is enabled; the default install stays read-only.
-  # patch on workloads/pods backs the annotate action; patch on nodes backs the
-  # cordon action; networkpolicies back the quarantine action.
+  # Provisioned once for the whole action set so later phases need not reopen
+  # this ClusterRole:
+  #   - patch on apps workloads + pods backs annotate (active now);
+  #   - patch on nodes backs cordon (later phase) -- cluster-wide, as Kubernetes
+  #     has no field-level RBAC for spec.unschedulable;
+  #   - networkpolicies backs quarantine (later phase).
   - apiGroups: ["apps"]
     resources: ["deployments", "statefulsets", "daemonsets"]
     verbs: ["patch"]
--- a/charts/kubescape-operator/templates/operator/clusterrole.yaml
+++ b/charts/kubescape-operator/templates/operator/clusterrole.yaml
@@ -46,4 +46,19 @@ rules:
     resources: ["securityexceptions", "clustersecurityexceptions"]
     verbs: ["get", "list", "watch"]
   {{- end }}
+  {{- if eq .Values.capabilities.remediation "enable" }}
+  # Post-scan remediation (opt-in). These mutating verbs are granted only when
+  # capabilities.remediation is enabled; the default install stays read-only.
+  # patch on workloads/pods backs the annotate action; patch on nodes backs the
+  # cordon action; networkpolicies back the quarantine action.
+  - apiGroups: ["apps"]
+    resources: ["deployments", "statefulsets", "daemonsets"]
+    verbs: ["patch"]
+  - apiGroups: [""]
+    resources: ["pods", "nodes"]
+    verbs: ["patch"]
+  - apiGroups: ["networking.k8s.io"]
+    resources: ["networkpolicies"]
+    verbs: ["create", "delete", "get", "list"]
+  {{- end }}
 {{- end }}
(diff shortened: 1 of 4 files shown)


matthyx:
Reviewed the chart change locally (rendered both ways, `helm lint` clean, `helm unittest` 28/28 + 966 snapshots green). **No blockers from my side** — the gating is correct, default-off is real, and the cross-component story checks out. A few non-blocking notes below.

**Verified**
- The `remediation` rule block is correctly nested inside `if $components.operator.enabled` and gated on `eq .Values.capabilities.remediation "enable"`; the default render stays read-only (the two focused unittests confirm present-when-enabled / absent-by-default).
- Cross-component safety holds: `operator`'s `LoadCapabilitiesConfig` uses a plain `viper.Unmarshal` (mapstructure, non-strict), so the new `remediation` key in the rendered `capabilities` configmap is silently ignored — an older operator image won't fail to start on it. Good to call this out explicitly in the PR body, which you did.

**Non-blocking observations**
1. **Phase-2/3 RBAC granted in a phase-1 PR.** `networkpolicies` (quarantine, phase 2) and `nodes` patch (cordon, phase 3) are provisioned now while only `annotate` is active. I'm fine with the "provision once so later PRs don't reopen the ClusterRole" rationale, but it does trade strict least-privilege-per-phase for convenience — worth a conscious sign-off rather than a silent default.
2. **`patch` on `nodes` is broad.** It's the minimal verb for cordon (there's no narrower RBAC for `spec.unschedulable`), but the same grant also allows mutating node labels/taints/spec cluster-wide. Gated + documented, so acceptable; just flagging the blast radius.
3. **`networkpolicies` verbs** are `create/delete/get/list` — sufficient for a create-on-quarantine / delete-on-release model, but no `update`/`watch`. If a later phase reconciles or watches existing policies it'll need revisiting. Fine for now.
4. **Minor:** the `values.yaml` comment says "Read the matching docs before enabling" but doesn't link them — consider pointing at the design doc / operator docs so operators have somewhere to land.

LGTM to merge once you're comfortable with (1).

matthyx (approved):

PR #864, opened 2026-06-22
feat(node-agent): expose alertDeduplication.bypass config value

slashben:
## What

Exposes `nodeAgent.config.alertDeduplication.bypass` (default `false`) and renders it into the node-agent ConfigMap as `"alertDeduplication": { "bypass": <bool> }`.

## Why

Lets operators toggle the node-agent alert-deduplication bypass (disables all sensor-side dedup/cooldown) via chart values, for high-volume security-testing (DAST) scenarios.

## Notes
- Default `false` → no behavior change.
- Snapshots regenerated; only the new block plus the expected node-agent config-checksum annotations changed.
- Pairs with the corresponding node-agent config switch.

## Summary by CodeRabbit

* **New Features**
  * Added a new node-agent option to control alert deduplication behavior.
  * Included a default setting so deduplication remains enabled unless explicitly turned off.


slashben:
Implements the config value added in kubescape/node-agent#838.

matthyx (approved):

PR #868, opened 2026-06-29
feat(node-agent): target label-less nodes via DoesNotExist affinity

matthyx:
## Problem

The node-agent autoscaler creates a DaemonSet per node group, targeting nodes with a `nodeSelector` on the grouping label. Nodes that don't carry the label (on-prem / custom clusters) can't be matched by any value-based selector, so they get no node-agent.

## Change

The operator now forms a **default group** for label-less nodes (`IsDefaultGroup`). This PR adds the matching DaemonSet template branch:

- The instance-type `nodeSelector` entry is guarded by `{{- if not .IsDefaultGroup }}`.
- For the default group, a `nodeAffinity` requires the grouping label to **`DoesNotExist`**, AND-ed into the same matchExpressions term as the existing `kubernetes.io/os=linux` requirement (one term ⇒ AND; separate terms would wrongly OR).
- Non-default groups render exactly as before.
- Snapshot regenerated — only the 2 autoscaler `template-configmap` snapshots changed; `helm unittest` 35/35 passing.

## Companion PR

Requires the operator change: kubescape/operator#389

🤖 Generated with [Claude Code](https://claude.com/claude-code)


matthyx (commented):
## Review — no blockers found ✅

Reviewed the DaemonSet template branch for the label-less "default group". The two-layer templating (Helm escaping `{{` `}}` so the operator sees literal Go-template directives) is correct, and I verified the operator-side render produces **valid YAML with the right semantics** for both group types:

**Non-default group** (`IsDefaultGroup=false`):
```yaml
nodeSelector:
  kubernetes.io/os: linux
  node.kubernetes.io/instance-type: "<group>"
affinity:
  nodeAffinity:
    requiredDuringSchedulingIgnoredDuringExecution:
      nodeSelectorTerms:
      - matchExpressions:
        - {key: kubernetes.io/os, operator: In, values: [linux]}
```

**Default group** (`IsDefaultGroup=true`):
```yaml
nodeSelector:
  kubernetes.io/os: linux          # instance-type selector correctly dropped
affinity:
  nodeAffinity:
    requiredDuringSchedulingIgnoredDuringExecution:
      nodeSelectorTerms:
      - matchExpressions:
        - {key: kubernetes.io/os, operator: In, values: [linux]}
        - {key: node.kubernetes.io/instance-type, operator: DoesNotExist}
```

Confirmed:
- Both `DoesNotExist` and the `os=linux` requirement land in the **same** `matchExpressions` term ⇒ AND-ed (a separate term would have OR-ed them and matched every linux node — the comment calls this out correctly).
- `{{- ... }}` left-trim cleanly removes the helper indentation so the emitted YAML stays correctly indented in both branches.
- The `DoesNotExist` key and the value-based `nodeSelector` key both derive from `nodeAgent.autoscaler.nodeGroupLabel`, so they stay in sync.
- Dropping `(not .autoscalerMode)` from the `customScheduling.affinity` else-if is safe — that branch is only reachable when not in autoscaler mode.
- Matches the operator contract in kubescape/operator#389 (`IsDefaultGroup`, default group name `"default"`, `DoesNotExist` affinity + no instance-type selector).
- All CI checks green; only the 2 autoscaler `template-configmap` snapshots changed.

### Non-blocking note
In autoscaler mode the affinity branch now **unconditionally** emits the operator-managed `nodeAffinity` and no longer falls through to a user-set `nodeAgent.affinity` / `customScheduling.affinity` (previously, `nodeAgent.affinity` would win even in autoscaler mode). This looks intentional — these DaemonSets are operator-managed and a custom affinity would break group targeting — but it is a silent behavior change for anyone who set `nodeAgent.affinity` alongside the autoscaler. Worth a one-line mention in release notes if that combination was ever supported.

Good to merge alongside operator#389.

matthyx:
🔗 Companion PR: kubescape/operator#389 — the operator change that forms the default group (`IsDefaultGroup`) consumed by this template branch.

slashben:
**Autoscaler mode now silently ignores user-set `nodeAgent.affinity`**

The affinity branch changed from `{{- if .Values.nodeAgent.affinity }}` to `{{- if .autoscalerMode }}` (with `.autoscalerMode` checked first). Previously, in autoscaler mode a user-overridden `nodeAgent.affinity` was rendered; now it is unconditionally replaced by the hardcoded `kubernetes.io/os In linux` term (plus the `DoesNotExist` expression for the default group).

For the shipped default (`nodeAgent.affinity` = the `os=linux` block) this is equivalent — which is why only the 2 autoscaler snapshots moved — but any autoscaler user who customized `nodeAgent.affinity` (zone / GPU / topology constraints) loses it with no warning. If operator-managed affinity is intended to be authoritative in autoscaler mode that's a reasonable call, but it's an undocumented behavior change worth a note in the values docs / PR description.

(Dropping `(not .autoscalerMode)` from the `customScheduling.affinity` branch is correct — that branch is unreachable in autoscaler mode now.)

matthyx:
Thanks @slashben — fixed in `89552893`.

Restructured the affinity block so only the **default** group (label-less nodes) gets the operator-managed `DoesNotExist` affinity; **non-default** autoscaler groups no longer get an operator-managed affinity at all and keep honouring user-provided `nodeAgent.affinity` / `customScheduling.affinity` (their OS requirement is already enforced by the nodeSelector). So a custom zone/GPU/topology affinity is preserved in autoscaler mode again. Added a note in `values.yaml` that the default group is the exception. Snapshot regenerated.

slashben (approved):

PR #869, opened 2026-06-29
feat(synchronizer): sync containerprofiles to backend

shanyl9:
## What

Adds `containerprofiles` to the in-cluster synchronizer's synced-resource list (strategy: `copy`), alongside `applicationprofiles` and `networkneighborhoods`.

## Changes
- **`templates/synchronizer/configmap.yaml`** — add `containerprofiles` (`spdx.softwarecomposition.kubescape.io/v1beta1`, strategy `copy`) to the synchronizer's resources, inside the `storage.enabled` block.
- **`templates/synchronizer/clusterrole.yaml`** — grant the synchronizer ServiceAccount RBAC on `containerprofiles` (`get/watch/list/create/update/patch/delete`), matching the existing AP/NN rule. Without this the synchronizer cannot read CPs and sync fails.
- **`tests/__snapshot__/snapshot_test.yaml.snap`** — regenerated via `helm-unittest`.

## Why
ContainerProfile storage has two mutually-exclusive modes:
- **Backend storage** — CPs are handled by the backend consolidator pipeline.
- **In-cluster storage** (`storage.enabled`) — profile CRDs live in the cluster and are synced to the backend by the synchronizer.

In the in-cluster-storage mode, `containerprofiles` was the only `spdx.softwarecomposition` profile CRD **not** in the synchronizer's resource list — `applicationprofiles` and `networkneighborhoods` were synced, but CPs were not. This closes that gap so CPs are synced like their AP/NN siblings.

The addition is inside the `storage.enabled` block, so it only takes effect in the in-cluster-storage deployment — it never runs alongside the backend consolidator path, so there is no duplicate handling.

## Testing
`docker run --rm -v "$(pwd)":/apps helmunittest/helm-unittest charts/kubescape-operator/` -> 37 tests, 966 snapshots passing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/charts/kubescape-operator/templates/synchronizer/clusterrole.yaml
+++ b/charts/kubescape-operator/templates/synchronizer/clusterrole.yaml
@@ -28,7 +28,7 @@ rules:
     resources: ["networkpolicies", "ingresses"]
     verbs: ["get", "list", "watch"]
   - apiGroups: ["spdx.softwarecomposition.kubescape.io"]
-    resources: ["applicationprofiles", "networkneighborhoods"]
+    resources: ["applicationprofiles", "networkneighborhoods", "containerprofiles"]
     verbs: ["get", "watch", "list", "create", "update", "patch", "delete"]
   - apiGroups: ["spdx.softwarecomposition.kubescape.io"]
     resources: ["knownservers"]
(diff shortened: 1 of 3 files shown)


matthyx (approved):
thanks @shanyl9

PR #872, opened 2026-07-03
fix(storage): restart storage pod when storage config changes

### LINKED ISSUE #867: Kubescape storage does not automatically pick up storage config changes
# Description
When updating storage config options (e.g. `storage.cleanupInterval`)  for an existing install, and absent other changes (e.g. version upgrade), the config changes do not take effect until the storage deployment is restarted.

# Environment
OS: Amazon Linux 2023
Version: kubescape-operator chart 1.40.1, quay.io/kubescape/storage:v0.0.274 
 
# Steps To Reproduce
1. Install kubescape-operator from helm chart with storage enabled
2. Upgrade helm release, same chart version, changing only `storage.cleanupInterval` input value

# Expected behavior
Kubescape storage honors the new storage config options (either by hot reload or automatic restart).

# Actual Behavior
Kubescape storage keeps using old config until restarted explicitly or unrelated reasons (pod failure, eviction, etc.).

# Additional context

I don't know if kubescape storage itself is capable of hot-reloading config file, but even if it is, it [will never see the updated config](https://kubernetes.io/docs/concepts/storage/volumes/#configmap).
> A container using a ConfigMap as a subPath volume mount will not receive updates when the ConfigMap changes.

matthyx:
## Overview
`storage/configmap.yaml` (`config.json`, e.g. `cleanupInterval`) is mounted into the kubescape-storage pod via a `subPath` volume mount, so kubelet never propagates ConfigMap updates to the running container. On top of that, the storage `Deployment` had no checksum annotation to trigger a rollout on `helm upgrade` either, so changing a value like `storage.cleanupInterval` never takes effect until the pod is restarted for an unrelated reason.

This adds a `storageConfig` checksum to the shared `checksums` helper (hash of `storage/configmap.yaml`) and wires it into the storage `Deployment`'s pod annotations, matching the pattern already used by `operator`, `node-agent`, and `synchronizer`. Also adds the `cloud-config` checksum, which was missing from this deployment as well (present on every other deployment).

## Related issues/PRs
Resolved #867

## How to Test
```
helm template charts/kubescape-operator --set account=<guid> --set storage.cleanupInterval=72h | grep checksum/storage-config
```
Changing `storage.cleanupInterval` (or any other `storage.*` config field) between installs now changes the annotation value, which triggers a pod restart on `helm upgrade`.

- [x] New and existing unit tests pass locally with my changes (`helm unittest charts/kubescape-operator`, 45/45 passing)

## Summary by CodeRabbit

* **Bug Fixes**
  * Storage component pods now include additional rollout checksums, helping them restart automatically when related configuration changes.
  * Pod annotations are now consistently rendered, even when only checksum-based updates are present.
* **Tests**
  * Updated Helm test coverage to include the storage configuration resource.

### DIFF
--- a/charts/kubescape-operator/templates/_common.tpl
+++ b/charts/kubescape-operator/templates/_common.tpl
@@ -11,6 +11,7 @@ matchingRulesConfig: {{ include (printf "%s/%s/%s" $.Template.BasePath $.Values.
 nodeAgentConfig: {{ include (printf "%s/node-agent/configmap.yaml" $.Template.BasePath) . | replace .Chart.AppVersion "" | sha256sum }}
 operatorConfig: {{ include (printf "%s/operator/configmap.yaml" $.Template.BasePath) . | replace .Chart.AppVersion "" | sha256sum }}
 proxySecret: {{ include (printf "%s/%s/%s" $.Template.BasePath $.Values.global.proxySecretDirectory "proxy-secret.yaml") . | replace .Chart.AppVersion "" | sha256sum }}
+storageConfig: {{ include (printf "%s/storage/configmap.yaml" $.Template.BasePath) . | replace .Chart.AppVersion "" | sha256sum }}
 synchronizerConfig: {{ include (printf "%s/synchronizer/configmap.yaml" $.Template.BasePath) . | replace .Chart.AppVersion "" | sha256sum }}
 admissionCertgenScripts: {{ include (printf "%s/operator/admission-webhook/configmap.yaml" $.Template.BasePath) . | replace .Chart.AppVersion "" | sha256sum }}
 storageCertgenScripts: {{ include (printf "%s/storage/certgen/configmap.yaml" $.Template.BasePath) . | replace .Chart.AppVersion "" | sha256sum }}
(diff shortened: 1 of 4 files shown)


matthyx:
## Review — no blockers found

Reviewed the change and validated it end-to-end. This is a correct, well-scoped fix for #867.

**Correctness verified**
- The new `storageConfig` checksum in the `checksums` helper hashes `storage/configmap.yaml`, which is the ConfigMap (`{{ .Values.storage.name }}`) mounted into the apiserver container at `/etc/config/config.json` via a `subPath` mount — so kubelet never propagates in-place updates and a checksum-triggered rollout is exactly the right mechanism.
- The added `checksum/cloud-config` (`cloudConfig` → `configs/cloudapi-configmap.yaml`) pairs correctly with the `.Values.global.cloudConfig` ConfigMap that is likewise mounted via `subPath` (`/etc/config/clusterData.json`). Same subPath propagation gap, same fix. This matches what operator/node-agent/synchronizer/kubescape/kubevuln already do.
- Rendered the chart with `storage.cleanupInterval=72h` vs `99h`: `checksum/storage-config` changes (`4002012b…` → `d9562c44…`) while `checksum/cloud-config` stays stable — confirming a `storage.*` change now triggers a storage-pod rollout on `helm upgrade`.

**Removed `{{- if or $podAnnotations .Values.storage.podAnnotations $hasCertgenChecksum }}` guard**
- Safe: the two checksum annotations are now always non-empty, so the `annotations:` block always has at least two entries and can never render empty/invalid YAML.

**Tests**
- `helm unittest .` → 37 tests / 966 snapshots passing, including the regenerated storage annotations in `snapshot_test.yaml.snap` and the added `storage/configmap.yaml` template in `operator_node_agent_autoscaler_test.yaml`.

**One thing to be aware of (not a blocker):** the storage `Deployment` uses `strategy: Recreate`, so the first `helm upgrade` onto this chart version will restart the storage (apiserver) pod once as the annotations get added, with a brief unavailability window during the recreate. That is a one-time, expected consequence of the fix.

LGTM.

PR #876, opened 2026-07-12
GKE Autopilot allowlist drift gate + helper scripts

slashben:
## What
Adds a **release-time drift gate** for the GKE Autopilot node-agent allowlist, plus the helper scripts it uses and a feature doc.

- `scripts/gke-allowlist/` — `allowlist-drift.py` (classifier), `check-drift.sh` (CI wrapper: renders the node-agent with all optional features on and checks it is a subset of the allowlist), `build-allowlists.py` / `refresh-digests.py` (allowlist-repo maintenance), and a README.
- `gke-allowlist/armo-kubescape-node-agent-1.40-v2.yaml` — vendored copy of the approved allowlist the gate checks against.
- `03-helm-release.yaml` — new `allowlist-drift-check` job gating the release.
- `docs/features/gke-autopilot-allowlist-tooling.md` — feature doc.

## Why
Autopilot admission is subset-based: if a chart change adds a node-agent env/mount/capability/container the approved allowlist doesn't list, customers can't install on Autopilot until a new allowlist is approved. The gate catches this before release.

## Bypass
If drift is intended (shipping a change that needs a new allowlist), open the Gerrit allowlist update, then re-run the release with **`BYPASS_ALLOWLIST_DRIFT: true`** — passes with a warning instead of failing.

## Validation
`check-drift.sh` run locally against the current chart + vendored allowlist → `no-op` (covered).

AI-skills: armosec-shared-rules:writing-docs,schedule,loop | cmds: /model,/login

## Summary by CodeRabbit

* **New Features**
  * Added a release-time gate to validate GKE Autopilot WorkloadAllowlist compatibility for Helm chart releases.
  * Added updated WorkloadAllowlist manifests for the node agent.
  * Added tooling to check allowlist drift, generate allowlists, and refresh approved image digests.
  * Introduced an optional override input to bypass drift failures with a warning.

* **Documentation**
  * Added a guide covering the allowlist tooling, drift behavior, local run instructions, and known check limitations.

PR #878, opened 2026-07-13
ci: run GKE Autopilot allowlist drift check on PRs

slashben:
## What

Runs the GKE Autopilot allowlist drift check on **pull requests**, not only at release. Adds `.github/workflows/pr-allowlist-drift.yaml`, which reuses the same `scripts/gke-allowlist/check-drift.sh` and vendored allowlist as the release gate.

Triggers on `pull_request` (`opened`, `reopened`, `synchronize`, `ready_for_review`, `labeled`, `unlabeled`) for PRs that touch `charts/kubescape-operator/**`, `gke-allowlist/**`, or `scripts/gke-allowlist/**`.

## Why

A chart change can add an env/mount/capability the approved `WorkloadAllowlist` doesn't cover — which silently breaks installs on GKE Autopilot until Google approves a new allowlist. Catching it at review time (not just at release) means the PR author is told before merge.

## Behavior

Unlike the release gate there's no `BYPASS_ALLOWLIST_DRIFT` input to type on a PR, so:

- **No drift** → check passes; any prior drift comment is removed.
- **Drift** → posts/updates a **sticky comment @-mentioning the PR author** with the missing fields + next step, and **fails** the check.
- **Acknowledged** → after opening the Gerrit allowlist update, a maintainer adds the **`allowlist-drift-ack`** label; the check re-runs on the label and passes ("set as pass after you opened Gerrit").
- **Crashed** (non-`0`/`2` exit) → hard-fails; the label does **not** bypass a crash.

Make it a required status in branch protection to block merge on un-acked drift.

## Notes

- **Stacked on #876** (`gke-allowlist-drift-ci`) — the check-drift script + vendored allowlist land there. Base is set to that branch; merge after #876. Rebase onto `main` once #876 merges.
- Uses only first-party actions (`actions/github-script`), no third-party comment action.
- **Fork PRs** get a read-only token and can't post the sticky comment — if this repo accepts fork PRs, move the comment step to a `workflow_run` companion. Documented in `docs/features/gke-autopilot-allowlist-tooling.md`.

AI-skills: armosec-shared-rules:writing-docs,schedule,loop | cmds: /model,/login,/compact

## Summary by CodeRabbit

* **New Features**
  * Added automated pull request checks to detect drift between the GKE Autopilot node-agent configuration and its vendored allowlist.
  * Checks post or update a status comment with results, including details when drift is detected.
  * Drift can be acknowledged with the designated label after the allowlist is updated.

* **Documentation**
  * Documented check triggers, outcomes, acknowledgements, failures, and fork pull request behavior.

### DIFF
--- a/.github/workflows/pr-allowlist-drift.yaml
+++ b/.github/workflows/pr-allowlist-drift.yaml
@@ -4,6 +4,12 @@ name: pr-allowlist-drift
 # BYPASS_ALLOWLIST_DRIFT input), a PR is automated — so on drift this notifies the PR author
 # with a sticky comment and fails the check until a maintainer adds the `allowlist-drift-ack`
 # label (the PR-flow equivalent of "set as pass after you opened Gerrit to allowlist").
+#
+# This check is advisory (a notification), not a required status. Do NOT mark it required in
+# branch protection while the `paths:` filter below is in place: GitHub skips a path-filtered
+# workflow on PRs that don't match, and a required check that never runs blocks the PR forever
+# on "Expected — waiting for status". To make it a hard gate, drop the paths filter and no-op
+# early inside the job instead.
 on:
   pull_request:
     types: [opened, reopened, synchronize, ready_for_review, labeled, unlabeled]
@@ -22,6 +28,9 @@ permissions:
 
 jobs:
   allowlist-drift:
+    # Only the `allowlist-drift-ack` label affects this check, so skip label events for any
+    # other label instead of re-running the whole checkout + helm + drift check needlessly.
+    if: ${{ !((github.event.action == 'labeled' || github.event.action == 'unlabeled') && github.event.label.name != 'allowlist-drift-ack') }}
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
@@ -50,18 +59,16 @@ jobs:
           set -e
           echo "$OUT"
           # rc: 0 = OK, 2 = spec-drift (ack-able), anything else = the check crashed.
+          # Pass the report through base64 so an arbitrary multiline blob can't break the
+          # step output (no heredoc delimiter to collide with / inject through).
           echo "rc=$rc" >> "$GITHUB_OUTPUT"
-          {
-            echo "report<<DRIFT_EOF"
-            echo "$OUT"
-            echo "DRIFT_EOF"
-          } >> "$GITHUB_OUTPUT"
+          echo "report_b64=$(printf '%s' "$OUT" | base64 -w0)" >> "$GITHUB_OUTPUT"
 
       - name: Notify author & conclude
         uses: actions/github-script@v7
         env:
           RC: ${{ steps.drift.outputs.rc }}
-          REPORT: ${{ steps.drift.outputs.report }}
+          REPORT_B64: ${{ steps.drift.outputs.report_b64 }}
           ACK_LABEL: allowlist-drift-ack
           ALLOWLIST: armo-kubescape-node-agent-1.40-v2
         with:
@@ -70,6 +77,7 @@ jobs:
             const rc = process.env.RC;
             const ackLabel = process.env.ACK_LABEL;
             const allowlist = process.env.ALLOWLIST;
+            const report = Buffer.from(process.env.REPORT_B64 || '', 'base64').toString('utf8').trim();
             const { owner, repo } = context.repo;
             const pr = context.payload.pull_request;
             const issue_number = pr.number;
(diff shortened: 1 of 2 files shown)


rotemamsa (dismissed):

.github/workflows/pr-allowlist-drift.yaml
```diff
@@ -0,0 +1,127 @@
+name: pr-allowlist-drift
+# Runs the GKE Autopilot allowlist drift check on every PR that touches the node-agent
+# render, the vendored allowlist, or the drift tooling. Unlike the release gate (a manual
+# BYPASS_ALLOWLIST_DRIFT input), a PR is automated — so on drift this notifies the PR author
+# with a sticky comment and fails the check until a maintainer adds the `allowlist-drift-ack`
+# label (the PR-flow equivalent of "set as pass after you opened Gerrit to allowlist").
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, ready_for_review, labeled, unlabeled]
+    paths:
```
matthyx, line 16: **`paths:` filter conflicts with "make it a required status".** The PR description recommends making this a required check in branch protection, but with a `paths:` filter GitHub will skip the workflow entirely on PRs that don't touch `charts/kubescape-operator/**` / `gke-allowlist/**` / `scripts/gke-allowlist/**`. A *required* check that never runs never reports a status, so those PRs sit permanently on "Expected — waiting for status to be reported" and can't be merged.

If this is meant to be required, either drop the `paths:` filter and no-op early inside the job, or add a companion job (no path filter) that always reports success on the non-matching case. Otherwise keep it non-required and call that out in the docs.

.github/workflows/pr-allowlist-drift.yaml
```diff
@@ -0,0 +1,127 @@
+name: pr-allowlist-drift
+# Runs the GKE Autopilot allowlist drift check on every PR that touches the node-agent
+# render, the vendored allowlist, or the drift tooling. Unlike the release gate (a manual
+# BYPASS_ALLOWLIST_DRIFT input), a PR is automated — so on drift this notifies the PR author
+# with a sticky comment and fails the check until a maintainer adds the `allowlist-drift-ack`
+# label (the PR-flow equivalent of "set as pass after you opened Gerrit to allowlist").
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, ready_for_review, labeled, unlabeled]
```
matthyx, line 15: **`labeled`/`unlabeled` fire on *every* label, not just the ack label.** Any label add/remove re-runs the full checkout + helm + python + drift check, even though only `allowlist-drift-ack` is relevant. It's wasteful rather than incorrect. Consider early-exiting in the github-script step when `github.event.action` is `labeled`/`unlabeled` and `github.event.label.name !== ACK_LABEL`.

.github/workflows/pr-allowlist-drift.yaml
```diff
@@ -0,0 +1,127 @@
...
+              '',
+              'On GKE Autopilot the node-agent would be rejected by GKE Warden until Google approves a matching allowlist, so customers could not install this chart version.',
+              '',
+              '**To unblock this PR:**',
+              '1. Open a Gerrit allowlist update at `gke-ap-allowlist.googlesource.com/ARMO` (see `scripts/gke-allowlist/README.md`), then update the vendored `gke-allowlist/` copy once approved.',
+              '2. A maintainer adds the **`' + ackLabel + '`** label to this PR to acknowledge the drift and pass the check.',
+              '',
+              detail,
+            ].join('\n');
+            await upsert(body);
```
matthyx, line 135: **Acknowledged drift still leaves the scary sticky comment.** When `rc === '2'` and the PR is `acked`, this `upsert(body)` runs first (posting/keeping the "⚠️ drifts from the approved allowlist" comment) and then the check passes at line 123. The result is a green check sitting next to a red warning comment, which reads as contradictory. Consider replacing the body with an "acknowledged" note (or calling `clear()`) in the `acked` branch instead of leaving the drift warning up.

.github/workflows/pr-allowlist-drift.yaml
```diff
@@ -0,0 +1,127 @@
...
+          OUT=$(scripts/gke-allowlist/check-drift.sh \
+            charts/kubescape-operator \
+            gke-allowlist/armo-kubescape-node-agent-1.40-v2.yaml 2>&1)
+          rc=$?
+          set -e
+          echo "$OUT"
+          # rc: 0 = OK, 2 = spec-drift (ack-able), anything else = the check crashed.
+          echo "rc=$rc" >> "$GITHUB_OUTPUT"
+          {
+            echo "report<<DRIFT_EOF"
```
matthyx: **`GITHUB_OUTPUT` heredoc uses a fixed delimiter.** If `check-drift.sh` ever emits a line equal to `DRIFT_EOF`, this multiline output block breaks (and could even be used to inject extra outputs). Low risk given the controlled script, but using a unique/random delimiter (e.g. `report<<EOF_$RANDOM` captured into a variable) removes the footgun entirely.

matthyx:
@slashben approved and merged all the other PRs, up to you to fix the commented nits above...

slashben:
Thanks @matthyx — all four addressed in dd3d5b3:

1. **`paths:` filter vs required check** — good catch. Kept the check **advisory** rather than required (it matches the intent here: *notify* the PR author, not hard-block). Dropped the "make it a required status" recommendation from the doc and replaced it with a note explaining the path-filter/required-check deadlock and how to convert it to a real gate (drop the filter, no-op early in the job) if we ever want that.
2. **`labeled`/`unlabeled` on every label** — added a job-level `if` that skips label events unless the label is `allowlist-drift-ack`, so unrelated label churn no longer spins up checkout+helm+python.
3. **Acked drift still showing the ⚠️ comment** — the `acked` branch now replaces the body with an "✅ acknowledged" note (and returns before the warning `upsert`), so a green check no longer sits next to a red warning.
4. **`GITHUB_OUTPUT` heredoc delimiter** — removed the heredoc entirely; the report is now base64-encoded into a single-line output and decoded in github-script, so no delimiter can collide or inject.

The identical workflow already merged into `armosec/helm-charts` via #36, so I opened **armosec/helm-charts#37** with the same four fixes there. Ready for another look when you have a minute.

matthyx (approved):

PR #884, opened 2026-07-27
move extra node-agent configuration to extra

matthyx:
## Summary by CodeRabbit

- **New Features**
  - Added support for passing custom node-agent settings through to generated configuration.
  - Backend storage can be enabled through node-agent configuration.
  - Added tracer disable options to node-agent configuration.

- **Changes**
  - Malware detection now requires runtime detection to be enabled.
  - Updated warnings for malware detection requests without runtime detection.
  - Custom settings that conflict with chart-managed options are rejected.

- **Removed**
  - Removed host malware sensor, file-size, and host network sensor settings.


matthyx:
### Note on `nodeAgent.config.extra` and collision guards

During review iterations, adding a hardcoded `$managedKeys` collision guard to reject extra keys that duplicate built-in fields was considered. However, we intentionally decided against maintaining such a list:
1. **Maintenance toil & drift:** Hand-maintaining a duplicate list of 40+ built-in `config.json` fields inside the Helm template creates ongoing maintenance burden and easily drifts out of sync whenever fields are added or updated (as seen with the recent tracer flags in #906).
2. **Standard Helm pattern:** `extra` is explicitly documented as an escape hatch for downstream/private distributions (e.g., `armosec/private-node-agent`) to pass fields not exposed by name in this public chart (`hostMalwareSensorEnabled`, `hostNetworkSensorEnabled`, `backendStorageEnabled`). In line with common Helm conventions (`extraConfig`, `extraArgs`), we rely on this clear contract rather than a fragile template-level blacklist.
3. **Special handling for `backendStorageEnabled`:** Because `backendStorageEnabled` additionally drives scan capability gating in `_common.tpl`, it remains explicitly handled, type-validated (must be `bool`), and backwards-compatible with legacy `capabilities.backend-storage: enable`.

charts/kubescape-operator/templates/_common.tpl
```diff
@@ -22,10 +22,20 @@ storageCertgenScripts: {{ include (printf "%s/storage/certgen/configmap.yaml" $.
 {{- $createCloudSecret := (empty .Values.credentials.cloudSecret) -}}
 {{- $submit := not (empty .Values.server) -}}
 {{- $virtualCrds := not (empty .Values.storage.forceVirtualCrds) -}}
+{{- $backendStorage := false -}}
+{{- if hasKey .Values.nodeAgent.config.extra "backendStorageEnabled" -}}
```
jnathangreeg: **HIGH — hard render failure when `extra` is nil.**

`hasKey` requires a map. `extra: {}` in `values.yaml` covers the default, but a user who writes the key with nothing under it — ordinary YAML, and what `--set nodeAgent.config.extra=null` produces — gets `nil`, and this line throws.

Reproduced:

```
$ helm template rel ./charts/kubescape-operator --set nodeAgent.config.extra=null ...
Error: template: kubescape-operator/templates/storage/certgen/rolebinding.yaml:1:29:
  executing "kubescape-operator/templates/storage/certgen/rolebinding.yaml" at <include "components" .>:
  ... executing "configurations" at <.Values.nodeAgent.config.extra>:
  wrong type for value; expected map[string]interface {}; got interface {}
```

Note it surfaces on `storage/certgen/rolebinding.yaml` — a template with nothing to do with the node-agent — because `configurations` is included everywhere. The user gets a raw Go type error pointing at an unrelated file.

**Fix:**

```
{{- $extra := default dict .Values.nodeAgent.config.extra -}}
{{- if hasKey $extra "backendStorageEnabled" -}}
```

The same guard is needed at `node-agent/configmap.yaml:30`, where `range` over nil is harmless but the value should come from the same expression.

charts/kubescape-operator/templates/node-agent/configmap.yaml
```diff
@@ -21,16 +21,26 @@ metadata:
 data:
   config.json: |
     {
+        {{- /*
+        nodeAgent.config.extra is an unvalidated passthrough intended for downstream/private
+        consumers to set fields not exposed by this public chart (e.g. hostNetworkSensorEnabled).
+        We intentionally avoid guarding against collisions with built-in keys here to prevent
+        the maintenance burden and drift of keeping a duplicate list of chart-managed keys.
+        */}}
+        {{- range $k, $v := .Values.nodeAgent.config.extra }}
```
jnathangreeg: **HIGH — a colliding key silently loses, and the losing side is the operator's.**

I've read @matthyx's note above and I'm **not** asking for a `$managedKeys` blacklist — that reasoning is sound and I agree a hand-maintained list of 40+ fields would drift. This is about ordering, which needs no list at all.

`extra` is emitted *before* the chart-managed keys, so a collision produces duplicate JSON member names and **Go's `encoding/json` takes the last occurrence** — the chart-managed value. The escape hatch loses to the thing it is supposed to escape.

Reproduced with `--set nodeAgent.config.extra.httpDetectionEnabled=false`:

```
467:        "httpDetectionEnabled": false,     <- from extra
471:        "httpDetectionEnabled": true,      <- chart-managed, wins
```

The operator sets a value, gets no error, and the chart quietly overrides it.

Also worth reconciling: the PR description states conflicting custom settings *"are now rejected"*. They aren't — they're accepted and then discarded.

**Fix — move the `range` block to the end of the object**, after `"seccompProfileBackend"` etc. Last-wins then makes `extra` authoritative, which matches the documented contract ("fields not exposed by this public chart" — and, if someone does override a named one, the override is what they asked for). Zero maintenance, no list, and the `backendStorageEnabled` special case at lines 31/35-37 is unaffected.

If instead the intent really is that chart-managed keys win, then the `values.yaml` comment and the PR description should say so explicitly, because the current wording implies the opposite.

charts/kubescape-operator/templates/_common.tpl
```diff
@@ -114,6 +124,12 @@ Because the gate logic lives here only, the rendered value and the warning about
 that value can never drift apart. See issue #851.
 */}}
 {{- define "capabilities.gates" -}}
+{{- if and (hasKey .Values.nodeAgent.config "hostMalwareSensor") (ne (get .Values.nodeAgent.config "hostMalwareSensor") "") }}
```
jnathangreeg: **MEDIUM/HIGH — this breaks `helm upgrade` for users carrying the shipped 1.40.3 default.**

The guard fires on **any** non-empty `hostMalwareSensor`, including `"disable"`. But `hostMalwareSensor: disable` is the value **released 1.40.3 ships as its own default** (`values.yaml:628` in the published tarball). Anyone who ran `helm show values kubescape-operator --version 1.40.3 > my-values.yaml` — the standard workflow — and kept the defaults now hits a hard render failure on upgrade, telling them to migrate a setting they never chose.

Reproduced:

```
$ helm template rel ./charts/kubescape-operator --set nodeAgent.config.hostMalwareSensor=disable ...
Error: execution error at (kubescape-operator/templates/synchronizer/deployment.yaml:1:28):
  nodeAgent.config.hostMalwareSensor has been removed. ...
```

Two reasons this is worth softening:

1. **The asymmetry is arbitrary.** Line 130 *tolerates* `hostNetworkSensor: disable` — the same "carried-over legacy default" case — and only fails on other values. Line 127 should do the same.
2. **`hostMalwareSensor: disable` is genuinely inert.** With the key removed, the agent defaults `hostMalwareSensorEnabled` to `false` anyway (`node-agent/pkg/config/config.go:172`). Failing on a value whose effective behaviour is identical to the new default buys nothing.

**Fix:** tolerate `"disable"` on line 127 the way line 130 already does, and keep failing on `"enable"` (which *would* change behaviour and genuinely needs migrating).

Minor converse, same pair of guards: `hostNetworkSensor: ""` currently **fails** with "has been removed", even though empty is inert. Reproduced. Accepting `""` and `"disable"` on both lines covers every no-op case.

charts/kubescape-operator/values.yaml
```diff
@@ -650,27 +650,21 @@ nodeAgent:
     syslogExporterURL: ""
     skipKernelVersionCheck: false
     malwareScanAllFiles: false
```
jnathangreeg, line 652: **MEDIUM — `hostMalwareSensorMaxFileSizeBytes` was removed with no guard, unlike the other two keys.**

This hunk drops three settings, but only `hostMalwareSensor` and `hostNetworkSensor` got `fail` guards in `_common.tpl:127-132`. The third is silently ignored.

Reproduced — set it and it vanishes with no error at all:

```
$ helm template rel ./charts/kubescape-operator     --set nodeAgent.config.hostMalwareSensorMaxFileSizeBytes=1048576 ... | grep -c hostMalwareSensorMaxFileSizeBytes
0
```

The consequence is worse than for the other two, because of the direction the default falls. The private agent reads this key (`private-node-agent/pkg/privateconfig/privateconfig.go`, `HostMalwareSensorMaxFileSizeBytes`), where **`0` means no cap** — the comment there notes an executed file is hashed whatever its size. So a user who had deliberately capped the sensor doesn't get an error telling them to move it to `extra`; they get **unbounded hashing on busy nodes**, discovered as CPU cost.

That is exactly the silent-loss failure the other two guards were added to prevent. Add a third guard alongside them:

```
{{- if hasKey .Values.nodeAgent.config "hostMalwareSensorMaxFileSizeBytes" }}
{{- fail "nodeAgent.config.hostMalwareSensorMaxFileSizeBytes has been removed. Use nodeAgent.config.extra.hostMalwareSensorMaxFileSizeBytes for custom/private node-agent builds." }}
{{- end }}
```

(If you'd rather not fail on the old default of `0`, gate it on a non-zero value — same shape as the `"disable"` tolerance I suggested on line 127.)

jnathangreeg (commented):
## The design is sound, and I verified the pair renders correctly together

I reviewed this alongside the downstream consumer it unblocks (armosec/helm-charts#39). To check the contract end to end I vendored **this branch's chart** into the armosec wrapper and rendered its values against it:

```
"hostMalwareSensorEnabled": false,
"hostNetworkSensorEnabled": false,
"backendStorageEnabled": true,
...
"effectiveCapabilities":{...,"nodeScan":"backend",...}
```

Exactly right, including the legacy `capabilities.backend-storage` compatibility path. The `extra` escape hatch does what it says, and moving the private-only sensor flags out of the public chart is the correct call — I confirmed the OSS agent declares `EnableHostMalwareSensor`/`EnableHostNetworkSensor` in `pkg/config/config.go` and never uses them; only `armosec/private-node-agent` consumes them.

Four things I'd fix before this is released, posted inline. Two are reproducible render failures.

### Also worth a look, not inline

- **`malwareDetectionEnabled` is newly ANDed with `runtimeDetection`** (`_common.tpl:142`), where it was previously the ungated `eq .Values.capabilities.malwareDetection "enable"`. The OSS agent branches on this flag standalone at `node-agent/cmd/main.go:355` (and treats it as an independent trigger at `:254`), so a user running `malwareDetection: enable` with `runtimeDetection: disable` **loses malware detection on upgrade**. Deliberate per the PR description, and defensible — but it is a silent narrowing whose only surfacing is a warning string. Nothing in `values.yaml` or an upgrade note says so.
- **`backendStorageEnabled` is now omitted when false** rather than emitted as `false`, unlike every other flag in the file. Harmless for viper (Go zero is false, and there's no `SetDefault` for it) but it changes `config.json` for every existing user, and anything asserting on the literal `"backendStorageEnabled": false` breaks.
- **Two coverage gaps.** The suite never exercises **`extra.backendStorageEnabled: true` on its own** (without `capabilities.backend-storage`) — which is precisely the path armosec#39 depends on. It covers the legacy capability, and `extra: false` overriding it, but not the new primary use case. Separately, the old `reports the capability as ineffective` assertion on `effectiveCapabilities.malwareDetection` has no replacement outside the snapshot. (CodeRabbit's note about covering an ordinary passthrough key is a different, also-valid gap.)

### The two deleted test values files are *not* a regression

`tests/values/malware_sensor_bool.yaml` and `malware_sensor_number.yaml` guarded the sprig-`default`-collapses-`false` trap on `hostMalwareSensor` — a hazard that ceases to exist once the key is gone. Equivalent bool/number/string type-rejection coverage is relocated onto `extra.backendStorageEnabled` at `malware_capability_gate_test.yaml:166` and `:181`. Coverage is preserved in substance; no action needed.

### Test results, honestly

`helm unittest` fails for me on **both** upstream `main` (175f824) and this branch: 16 tests / 24 snapshots, in the `Snapshot tests` suite only, with **identical counts and identical checksum deltas on both**. That is pre-existing local env drift (helm 3.15.3 / helm-unittest 0.6.1), **not introduced by this PR**. All 12 non-snapshot suites pass on both, including `Malware capability gate tests`.

### Ordering

#884 merges → a `kubescape-operator` chart release is cut → armosec/helm-charts#39 bumps its pins to it. Unless #39 takes the compatibility fix I suggested there (keep `capabilities.backend-storage: enable` alongside the new `extra` key, which this PR's legacy path still honours), in which case #39 can land independently and this PR is not on its critical path.

jnathangreeg (approved):

PR #890, opened 2026-07-29
feat: expose defaultFrameworks in kubescape-operator chart

### LINKED ISSUE #566: Enable Framework Selection During Helm Installation
## Overview
Currently, all frameworks are scanned by default, and there is no existing mechanism to control this behaviour via Helm.

## Problem
During installation, it is not possible to specify the framework required for a specific cloud provider. As a result, I must manually extract the necessary controls from the JSON reports, which can be a time-consuming process.

## Solution
Introduce a capability similar to Trivy Operator, allowing users to specify one or multiple frameworks in the `values.yaml` file. For example:

```bash
  specs:  
    - k8s-cis-1.23  
    - k8s-nsa-1.0  
    - k8s-pss-baseline-0.1  
    - k8s-pss-restricted-0.1  

Additionally, enable this configuration through the `--set` flag in the Helm install command.

## Additional Information
I would want to be able to set for GKE:
```bash
frameworks:
- AllControls
- cis-v1.23-t1.0.1

For AKS:
```bash
framworks:
- cis-aks-t1.2.0

### PR DESCRIPTION
**Pull request by manumathon**
## Overview
Add `defaultFrameworks` to the kubescape-operator chart and render it into `clusterData`, so users can pick posture frameworks at install time (e.g. `--set defaultFrameworks={cis-aks-t1.2.0}`).

## How to Test
```bash
helm template test . --set clusterName=x | grep -A2 defaultFrameworks
helm template test . --set clusterName=x --set defaultFrameworks={cis-aks-t1.2.0} | grep -A2 defaultFrameworks
```

## Related issues/PRs:
* Resolved #566
* Companion operator PR: https://github.com/kubescape/operator/pull/398

## Summary by CodeRabbit

* **New Features**
  * Added Helm configuration for defining default posture frameworks during installation.
  * Scheduler scans now use these defaults when no specific frameworks are requested.
  * Explicit scan targets continue to override configured defaults.
  * Framework entries are trimmed and validated, with empty values omitted.

* **Documentation**
  * Updated chart value and scheduler configuration documentation with setup details, fallback behavior, and override options.

### DIFF
--- a/charts/kubescape-operator/templates/configs/cloudapi-configmap.yaml
+++ b/charts/kubescape-operator/templates/configs/cloudapi-configmap.yaml
@@ -56,9 +56,20 @@ data:
 {{- end }}
       "relevantImageVulnerabilitiesConfiguration": "{{ .Values.capabilities.relevancy }}",
       "riskAcceptance": {{ eq .Values.capabilities.riskAcceptance "enable" }}
-      {{- if .Values.defaultFrameworks }},
-      "defaultFrameworks": {{ .Values.defaultFrameworks | toJson }}
-      {{- end }}
+{{- if .Values.defaultFrameworks }}
+{{- if not (kindIs "slice" .Values.defaultFrameworks) }}
+{{- fail (printf "defaultFrameworks must be a list of framework names, got %s (use --set 'defaultFrameworks={nsa,mitre}')" (kindOf .Values.defaultFrameworks)) }}
+{{- end }}
+{{- $frameworks := list }}
+{{- range .Values.defaultFrameworks }}
+{{- if and . (ne (trim . | toString) "") }}
+{{- $frameworks = append $frameworks . }}
+{{- end }}
+{{- end }}
+{{- if $frameworks }},
+      "defaultFrameworks": {{ $frameworks | toJson }}
+{{- end }}
+{{- end }}
       {{- if .Values.kubevuln.config.proxyRegistryMap }},
       "proxyRegistryMap": {{ toJson .Values.kubevuln.config.proxyRegistryMap }}
       {{- end }}
--- a/charts/kubescape-operator/templates/configs/cloudapi-configmap.yaml
+++ b/charts/kubescape-operator/templates/configs/cloudapi-configmap.yaml
@@ -56,6 +56,9 @@ data:
 {{- end }}
       "relevantImageVulnerabilitiesConfiguration": "{{ .Values.capabilities.relevancy }}",
       "riskAcceptance": {{ eq .Values.capabilities.riskAcceptance "enable" }}
+      {{- if .Values.defaultFrameworks }},
+      "defaultFrameworks": {{ .Values.defaultFrameworks | toJson }}
+      {{- end }}
       {{- if .Values.kubevuln.config.proxyRegistryMap }},
       "proxyRegistryMap": {{ toJson .Values.kubevuln.config.proxyRegistryMap }}
       {{- end }}
--- a/charts/kubescape-operator/templates/configs/cloudapi-configmap.yaml
+++ b/charts/kubescape-operator/templates/configs/cloudapi-configmap.yaml
@@ -56,14 +56,18 @@ data:
 {{- end }}
       "relevantImageVulnerabilitiesConfiguration": "{{ .Values.capabilities.relevancy }}",
       "riskAcceptance": {{ eq .Values.capabilities.riskAcceptance "enable" }}
-{{- if .Values.defaultFrameworks }}
+{{- if not (kindIs "invalid" .Values.defaultFrameworks) }}
 {{- if not (kindIs "slice" .Values.defaultFrameworks) }}
 {{- fail (printf "defaultFrameworks must be a list of framework names, got %s (use --set 'defaultFrameworks={nsa,mitre}')" (kindOf .Values.defaultFrameworks)) }}
 {{- end }}
 {{- $frameworks := list }}
 {{- range .Values.defaultFrameworks }}
-{{- if and . (ne (trim . | toString) "") }}
-{{- $frameworks = append $frameworks . }}
+{{- if not (kindIs "string" .) }}
+{{- fail (printf "defaultFrameworks entries must be strings, got %s" (kindOf .)) }}
+{{- end }}
+{{- $name := trim . }}
+{{- if ne $name "" }}
+{{- $frameworks = append $frameworks $name }}
 {{- end }}
 {{- end }}
 {{- if $frameworks }},
--- a/charts/kubescape-operator/tests/default_frameworks_test.yaml
+++ b/charts/kubescape-operator/tests/default_frameworks_test.yaml
@@ -52,3 +52,34 @@ tests:
     asserts:
       - failedTemplate:
           errorMessage: "defaultFrameworks must be a list of framework names, got string (use --set 'defaultFrameworks={nsa,mitre}')"
+
+  - it: should fail when defaultFrameworks is a falsy non-list
+    set:
+      unittest: true
+      clusterName: kind-kind
+      defaultFrameworks: false
+    asserts:
+      - failedTemplate:
+          errorMessage: "defaultFrameworks must be a list of framework names, got bool (use --set 'defaultFrameworks={nsa,mitre}')"
+
+  - it: should fail when an entry is not a string
+    set:
+      unittest: true
+      clusterName: kind-kind
+      defaultFrameworks:
+        - 1
+    asserts:
+      - failedTemplate:
+          errorMessage: "defaultFrameworks entries must be strings, got float64"
+
+  - it: should trim framework names
+    set:
+      unittest: true
+      clusterName: kind-kind
+      defaultFrameworks:
+        - "  nsa  "
+        - mitre
+    asserts:
+      - matchRegex:
+          path: data.clusterData
+          pattern: '"defaultFrameworks":\s*\["nsa","mitre"\]'
--- a/charts/kubescape-operator/tests/default_frameworks_test.yaml
+++ b/charts/kubescape-operator/tests/default_frameworks_test.yaml
@@ -1,0 +1,33 @@
+suite: defaultFrameworks in clusterData
+templates:
+  - configs/cloudapi-configmap.yaml
+tests:
+  - it: should render default frameworks into clusterData
+    set:
+      unittest: true
+      clusterName: kind-kind
+    asserts:
+      - matchRegex:
+          path: data.clusterData
+          pattern: '"defaultFrameworks":\s*\["allcontrols","nsa","mitre"\]'
+
+  - it: should allow overriding default frameworks
+    set:
+      unittest: true
+      clusterName: kind-kind
+      defaultFrameworks:
+        - cis-aks-t1.2.0
+    asserts:
+      - matchRegex:
+          path: data.clusterData
+          pattern: '"defaultFrameworks":\s*\["cis-aks-t1.2.0"\]'
+
+  - it: should omit defaultFrameworks when empty
+    set:
+      unittest: true
+      clusterName: kind-kind
+      defaultFrameworks: []
+    asserts:
+      - notMatchRegex:
+          path: data.clusterData
+          pattern: '"defaultFrameworks"'
--- a/charts/kubescape-operator/tests/default_frameworks_test.yaml
+++ b/charts/kubescape-operator/tests/default_frameworks_test.yaml
@@ -2,14 +2,14 @@ suite: defaultFrameworks in clusterData
 templates:
   - configs/cloudapi-configmap.yaml
 tests:
-  - it: should render default frameworks into clusterData
+  - it: should omit defaultFrameworks when empty (chart default)
     set:
       unittest: true
       clusterName: kind-kind
     asserts:
-      - matchRegex:
+      - notMatchRegex:
           path: data.clusterData
-          pattern: '"defaultFrameworks":\s*\["allcontrols","nsa","mitre"\]'
+          pattern: '"defaultFrameworks"'
 
   - it: should allow overriding default frameworks
     set:
@@ -20,9 +20,9 @@ tests:
     asserts:
       - matchRegex:
           path: data.clusterData
-          pattern: '"defaultFrameworks":\s*\["cis-aks-t1.2.0"\]'
+          pattern: '"defaultFrameworks":\s*\["cis-aks-t1\.2\.0"\]'
 
-  - it: should omit defaultFrameworks when empty
+  - it: should omit defaultFrameworks when explicitly empty
     set:
       unittest: true
       clusterName: kind-kind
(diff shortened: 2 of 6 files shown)


charts/kubescape-operator/values.yaml
```diff
@@ -13,6 +13,16 @@ fullnameOverride: "kubescape"
 # --set clusterName=`kubectl config current-context`
 clusterName: "cluster"
 
+# Default posture frameworks used when a scan does not specify frameworks
+# (startup scan, empty scheduler scanV1, exception-triggered rescan).
+# Explicit API / scanV1 targetNames override this list.
+# Example (AKS): --set defaultFrameworks={cis-aks-t1.2.0}
+# Example (GKE): --set 'defaultFrameworks={allcontrols,cis-v1.23-t1.0.1}'
+defaultFrameworks:
```
matthyx: **BLOCKER** — shipping a non-empty default here changes the daily posture scan for every existing installation on `helm upgrade`, with no value edited.

The chart already ships `kubescapeScheduler.requestBody.commands[0].args.scanV1: {}`. Today the operator turns that into:

```go
// setDefaultsKubescapeScanRequest, before #398
if len(postScanRequest.TargetNames) == 0 {
    postScanRequest.TargetNames = []string{"all"}
```

So the CronJob currently scans **all** frameworks. With kubescape/operator#398 plus this default, the exact same CronJob becomes `["allcontrols","nsa","mitre"]` — every user upgrading silently loses `cis-*`, `SOC2`, and the rest from their scheduled scan. Verified rendering:

```
$ helm template t charts/kubescape-operator --set clusterName=x | ...
"defaultFrameworks": ["allcontrols","nsa","mitre"]
```

Note the #566 comment saying the default should be the trio "as it currently does" is only accurate for the *startup* scan and the exception rescan (`GetStartupActions` / `buildRescanCommand`, both hardcoded to the trio). It is **not** accurate for the empty-`scanV1` scheduler path this PR also reroutes, which is `"all"` today. That's the divergence that makes this a regression rather than a no-op.

Three ways out, pick one deliberately:

1. `defaultFrameworks: []` — pure opt-in, zero behaviour change. This also keeps #398's `"all"` legacy fallback and `nativeDefaultFrameworks` reachable, which is what that PR's "safe if the field is missing" note promises. My preference.
2. `defaultFrameworks: [all]` — preserves current scheduler behaviour but changes the startup/exception rescans from the trio to everything. Louder, not obviously better.
3. Keep the trio and treat it as an intentional breaking change — then it needs a release note saying scheduled scans narrow from all frameworks to three, and it should not ride along in a patch bump.

Whichever you choose: with a non-empty default shipped, `"all"` and `nativeDefaultFrameworks` in #398 become dead code for anyone on this chart version. Worth being explicit about that in one place rather than leaving two "defaults" that disagree.

Separately, the comment block should document how to actually turn this off. `[]` in a values file works, but neither `--set` form does — see the comment on the test file.

charts/kubescape-operator/templates/configs/cloudapi-configmap.yaml
```diff
@@ -56,6 +56,9 @@ data:
 {{- end }}
       "relevantImageVulnerabilitiesConfiguration": "{{ .Values.capabilities.relevancy }}",
       "riskAcceptance": {{ eq .Values.capabilities.riskAcceptance "enable" }}
+      {{- if .Values.defaultFrameworks }},
+      "defaultFrameworks": {{ .Values.defaultFrameworks | toJson }}
```
matthyx: **BLOCKER** — `toJson` will faithfully render whatever type the user supplied, and `InstallationData.DefaultFrameworks` is `[]string`. A wrong type doesn't degrade this one field, it makes `json.Unmarshal` of the *whole* `clusterData` fail — and that same file is mounted at `/etc/config/clusterData.json` by operator, kubevuln, kubescape, node-agent, synchronizer and storage.

Both of these render happily today:

```
$ helm template ... --set defaultFrameworks=nsa
"defaultFrameworks": "nsa"          # JSON string, not a list

$ helm template ... --set defaultFrameworks.a=b
"defaultFrameworks": {"a": "b"}    # object
```

`--set defaultFrameworks=nsa` is exactly what a user will type first — the documented list form needs Helm's brace syntax (`--set defaultFrameworks={nsa}`), and nothing tells them so at render time.

There's precedent for the guard 48 lines up in this same template (the `cveMatchingMode` `fail`), so something like:

```gotemplate
{{- if and (.Values.defaultFrameworks) (not (kindIs "slice" .Values.defaultFrameworks)) }}
  {{- fail (printf "defaultFrameworks must be a list of framework names, got %s (use --set 'defaultFrameworks={nsa,mitre}')" (kindOf .Values.defaultFrameworks)) }}
{{- end }}
```

Worth also rejecting empty/blank entries in the list while you're there — `--set defaultFrameworks={}` renders `[""]`, which reaches the operator as a framework literally named `""`.

charts/kubescape-operator/tests/default_frameworks_test.yaml
```diff
@@ -0,0 +1,33 @@
...
+      unittest: true
+      clusterName: kind-kind
+      defaultFrameworks:
+        - cis-aks-t1.2.0
+    asserts:
+      - matchRegex:
+          path: data.clusterData
+          pattern: '"defaultFrameworks":\s*\["cis-aks-t1.2.0"\]'
+
+  - it: should omit defaultFrameworks when empty
```
matthyx: This case only proves the *values-file* form omits the key. None of the three `--set` spellings a user would reach for behave the way this test implies, and none are covered:

| command | rendered |
|---|---|
| `--set defaultFrameworks=[]` | `"defaultFrameworks": "[]"` — the two-character **string** `[]`, key still present, and an invalid type per the `cloudapi-configmap.yaml` comment |
| `--set defaultFrameworks={}` | `"defaultFrameworks": [""]` — one framework named empty string |
| `--set defaultFrameworks=null` | key omitted ✅ |
| `--set-json defaultFrameworks='[]'` | key omitted ✅ |

So `--set defaultFrameworks=null` is the answer, and it's the one spelling not mentioned anywhere in `values.yaml`. Please document it next to the AKS/GKE examples, and add a case here (or lean on the type guard to reject the first two loudly instead of silently).

matthyx (commented):
Thanks for picking this up — the shape is right (`defaultFrameworks` as a list, rendered into `clusterData`, explicit `targetNames` winning) and it matches the naming agreed in #566. Three blockers before this can go in, plus a couple of nits. Inline comments cover the rest.

### BLOCKER 1 — snapshot tests were not regenerated; CI will fail

Adding a key to `clusterData` changes the `ks-cloud-config` ConfigMap, which changes the `checksum/cloud-config` pod annotation on **every** workload that mounts it. Running the same command CI runs:

```
$ helm unittest charts/kubescape-operator/
Charts:      1 failed, 0 passed, 1 total
Test Suites: 1 failed, 4 passed, 5 total
Tests:       19 failed, 21 passed, 40 total
Snapshot:    92 failed, 904 passed, 996 total
```

Affected templates: `operator/deployment.yaml`, `kubevuln/deployment.yaml`, `kubescape/deployment.yaml`, `node-agent/*` (incl. the embedded `daemonset-template.yaml` in `template-configmap.yaml`), `storage/deployment.yaml`, `synchronizer/deployment.yaml`.

Fix:

```bash
helm unittest charts/kubescape-operator/ -u
git add charts/kubescape-operator/tests/__snapshot__/snapshot_test.yaml.snap
```

Your two new/updated suites do pass in isolation (`tests/default_frameworks_test.yaml` 3/3, `tests/offline_mode_test.yaml` 4/4) — it's only the snapshot suite. Heads-up on why you haven't seen this: the `pull_request_created` workflow is sitting in `action_required` (first-time contributor), so the unittest job hasn't reported on this PR yet. It will fail as-is once a maintainer approves the run.

### BLOCKER 2 — the default value silently narrows the scheduled scan for every existing user

See the inline comment on `values.yaml`. Short version: `scanV1: {}` + `"all"` today → `scanV1: {}` + the native trio after this PR. That is a behaviour change on plain `helm upgrade`, with no value edited.

### BLOCKER 3 — no type guard on the new value; a plausible `--set` typo produces a `clusterData` that no component can parse

See the inline comment on `cloudapi-configmap.yaml`.

### Nits

- `charts/kubescape-operator/README.md` — the values table isn't updated. Top-level keys (`clusterName`, `ksNamespace`, …) aren't in that table today either, so this is consistent as-is; but given the entire point of the feature is install-time ergonomics, a row would help discoverability.
- `tests/default_frameworks_test.yaml:23` — `\["cis-aks-t1.2.0"\]` leaves the dots unescaped in a regex. Harmless, matches anyway.
- No chart version bump needed here — this repo bumps `Chart.yaml` in separate `prepare release` PRs (79bd478, 7feaca6), so leaving 1.40.3 alone is correct.

### Merge order

As you noted in #566, kubescape/operator#398 has to ship first and the chart must point at an operator image that reads `defaultFrameworks`. Worth confirming that the operator image tag in `values.yaml` is bumped in the release PR that carries this, otherwise the key renders into `clusterData` and is ignored.

manumathon:
Thanks for the review — fixed the three blockers and the nits. Ready for another look.

charts/kubescape-operator/templates/configs/cloudapi-configmap.yaml
```diff
@@ -56,6 +56,20 @@ data:
 {{- end }}
       "relevantImageVulnerabilitiesConfiguration": "{{ .Values.capabilities.relevancy }}",
       "riskAcceptance": {{ eq .Values.capabilities.riskAcceptance "enable" }}
+{{- if .Values.defaultFrameworks }}
+{{- if not (kindIs "slice" .Values.defaultFrameworks) }}
+{{- fail (printf "defaultFrameworks must be a list of framework names, got %s (use --set 'defaultFrameworks={nsa,mitre}')" (kindOf .Values.defaultFrameworks)) }}
+{{- end }}
+{{- $frameworks := list }}
+{{- range .Values.defaultFrameworks }}
+{{- if and . (ne (trim . | toString) "") }}
```
matthyx: Two small follow-ups on this line, both non-blocking.

**1. `trim` runs before anything checks the element type**, so a non-string entry escapes the nice `fail` two lines up and produces a raw template error:

```
$ cat vals.yaml
defaultFrameworks:
  - 1

$ helm template t charts/kubescape-operator -f vals.yaml
Error: template: .../synchronizer/deployment.yaml:1:28: executing "..." at
<include "checksums" .>: ... at <.>: wrong type for value; expected string; got float64
```

Note it blames `synchronizer/deployment.yaml` (the `checksums` include pulls this template in), so the user gets pointed at a completely unrelated file. Same for a map element via `--set-json`. Cheap fix — validate inside the range and reuse the same message style:

```gotemplate
{{- range .Values.defaultFrameworks }}
{{- if not (kindIs "string" .) }}
  {{- fail (printf "defaultFrameworks entries must be strings, got %s" (kindOf .)) }}
{{- end }}
```

**2. entries are trimmed to decide, but appended untrimmed:**

```
defaultFrameworks: ["  nsa  ", mitre]
→ "defaultFrameworks": ["  nsa  ","mitre"]
```

The operator gets a framework name with leading/trailing spaces that will never match. You're already calling `trim` here, so `append $frameworks (trim .)` costs nothing.

charts/kubescape-operator/templates/configs/cloudapi-configmap.yaml
```diff
@@ -56,6 +56,20 @@ data:
 {{- end }}
       "relevantImageVulnerabilitiesConfiguration": "{{ .Values.capabilities.relevancy }}",
       "riskAcceptance": {{ eq .Values.capabilities.riskAcceptance "enable" }}
+{{- if .Values.defaultFrameworks }}
```
matthyx: Non-blocking, and I think this is the same point CodeRabbit raised: because the truthiness check is the outer gate, a non-list *falsy* scalar never reaches the `kindIs "slice"` guard below and is silently swallowed:

```
--set defaultFrameworks=false   → key omitted, no error
--set defaultFrameworks=0       → key omitted, no error
```

So `nsa` and `[]` fail loudly (good) but `false` quietly turns the feature off. Making the outer gate a nil check instead lets every non-list value fall through to the `fail`:

```gotemplate
{{- if not (kindIs "invalid" .Values.defaultFrameworks) }}
{{- if not (kindIs "slice" .Values.defaultFrameworks) }}
{{- fail (printf "defaultFrameworks must be a list ..." ...) }}
{{- end }}
```

An empty list is still falsy under `if $frameworks` further down, so the omit-when-empty behaviour and its test are unaffected.

charts/kubescape-operator/values.yaml
```diff
@@ -13,6 +13,16 @@ fullnameOverride: "kubescape"
...
 clusterName: "cluster"
 
+# Default posture frameworks used when a scan has no targetNames
+# (empty scheduler scanV1, startup scan, exception-triggered rescan).
+# Empty by default (opt-in) so upgrades do not narrow scheduled scans from "all".
+# Explicit scanV1/API targetNames override this list.
+# Clear with: --set defaultFrameworks=null
+# Example (AKS CIS): --set 'defaultFrameworks={cis-aks-t1.2.0}'
+# Example (allcontrols + current CIS): --set 'defaultFrameworks={allcontrols,cis-v1.10.0}'
+# Note: when triggerSecurityFramework is true (default), "security" is also scanned.
```
matthyx: Good addition and the substance checks out — I verified the chart sets `operator.triggerSecurityFramework: true` at line 431, which overrides the operator's own `viper.SetDefault("triggerSecurityFramework", false)`, so "true (default)" is correct *for chart users*.

Only nit: write it as `operator.triggerSecurityFramework` so someone who greps `values.yaml` for the bare name lands on the knob rather than just on this comment.

matthyx (commented):
Re-reviewed at 16a1c9a. **All three blockers are resolved**, and the fix for #2 was the better of the options — defaulting to `[]` means `clusterData` is byte-identical at chart defaults, so blocker #1 dissolved rather than needing a snapshot regeneration. Full suite is green:

```
$ helm unittest charts/kubescape-operator/
Charts:      1 passed, 1 total
Test Suites: 5 passed, 5 total
Tests:       42 passed, 42 total
Snapshot:    996 passed, 996 total
```

Re-verified the type guard by hand against the rendered output:

| command | before | now |
|---|---|---|
| _(chart default)_ | `["allcontrols","nsa","mitre"]` | key omitted ✅ |
| `--set 'defaultFrameworks={nsa}'` | `["nsa"]` | `["nsa"]` ✅ |
| `--set defaultFrameworks=nsa` | `"nsa"` (broke every component) | `fail` with the `{nsa,mitre}` hint ✅ |
| `--set defaultFrameworks.a=b` | `{"a":"b"}` | `fail`, `got map` ✅ |
| `--set defaultFrameworks=[]` | `"[]"` (silent string trap) | `fail`, `got string` ✅ |
| `--set defaultFrameworks={}` | `[""]` | key omitted ✅ |
| `--set defaultFrameworks=null` | omitted, undocumented | omitted, documented in `values.yaml` + README ✅ |
| `["", "nsa"]` | n/a | `["nsa"]` ✅ |

Also confirmed `offline_mode_test.yaml` is correctly reverted, the README row landed, the regex dots are escaped, and the suite grew from 3 to 5 cases including a `failedTemplate` assertion. Nice work.

Nothing blocking left. Four small things inline — take or leave them, none need another round from me:

- non-string list elements skip the friendly `fail` and surface a raw Go error blamed on the wrong template
- `--set defaultFrameworks=false` / `=0` are silently ignored instead of failing (CodeRabbit flagged this one too)
- entries are trimmed for the emptiness test but appended untrimmed
- one doc path nit on the new `triggerSecurityFramework` note

One thing to keep on the checklist for whoever cuts the release: kubescape/operator#398 still has to merge and be released first, and the chart's operator image tag bumped alongside — otherwise `defaultFrameworks` renders into `clusterData` and is ignored. With the default now `[]`, that ordering is no longer dangerous, just inert, which is a much better place to be.

matthyx (dismissed):

matthyx:
@manumathon I've released https://github.com/kubescape/operator/releases/tag/v0.2.161 can you bump it in that PR so the fix is complete?

manumathon:
@matthyx pushed the required changes, can you review the changes .

matthyx (approved):

PR #891, opened 2026-07-31
fix(node-agent): add OTEL_METRICS_EXPORTER env var for Prometheus exporter

### LINKED ISSUE #889: node-agent Prometheus metrics never start on :8080
# Description

Setting `nodeAgent.config.prometheusExporter: enable` doesn't actually turn on
node-agent's `/metrics` endpoint anymore. It still gets rendered into
`config.json` (`prometheusExporterEnabled: true`), but node-agent's metrics
pipeline moved to the OTEL SDK a while back, and the `:8080` scrape endpoint is
now gated behind an env var instead:

https://github.com/kubescape/node-agent/blob/6a86345e3b2c91635df705ab99ea14cd27bc02e3/pkg/otelsetup/setup.go#L64-L104

```go
if os.Getenv("OTEL_METRICS_EXPORTER") == "prometheus" {
    promShutdown, perr := initPrometheusMeterProvider(cfg)
    ...
}
return baseShutdown, nil // no metrics listener otherwise
```
It seems that the chart never sets OTEL_METRICS_EXPORTER anywhere. So config.prometheusExporter is a
not working as intended.

This used to work: at chart 1.30.6 (node-agent v0.3.79), pkg/otelsetup
didn't exist yet and the metrics server was a plain unconditional
http.ListenAndServe(":8080", nil). The gating was introduced somewhere in the
OTEL metrics migration between v0.3.79 and v0.3.158.

# Environment
OS: in-cluster kubescape-operator/node-agent (K8s version v1.34.8)
Version: kubescape-operator chart 1.40.3, node-agent v0.3.158 (chart's own default). Last known working on chart 1.30.6 / node-agent v0.3.79.
 
# Steps To Reproduce
1. Install kubescape-operator 1.40.3+ with capabilities.prometheusExporter: enable and nodeAgent.config.prometheusExporter: enable.
2. node-agent pods come up healthy, Service has endpoints on port 8080.
3. Scrape any node-agent pod on :8080/metrics — connection refused.

# Expected behavior
:8080/metrics works when nodeAgent.config.prometheusExporter: enable is set, like it did before 1.40.x.

# Actual Behavior
Nothing listens on 8080. Ruled out pods/Service/ServiceMonitor config, NetworkPolicy/CiliumNetworkPolicy, and CPU/memory throttling — the listener is just never started.

# Additional context
Workaround we're running with meanwhile:
nodeAgent:
  env:
    - name: NodeName
      valueFrom:
        fieldRef:
          fieldPath: spec.nodeName
    - name: OTEL_METRICS_EXPORTER
      value: prometheus

MihaiB-dev:
## Overview
It was added in `templates/node-agent/_node-agent.tpl` the env OTEL_METRICS_EXPORTER which is a must for `prometheusExporterEnabled`

## Additional Information

- Root cause: [`pkg/otelsetup/setup.go`](https://github.com/kubescape/node-agent/blob/6a86345e3b2c91635df705ab99ea14cd27bc02e3/pkg/otelsetup/setup.go#L64-L104)
  only calls `initPrometheusMeterProvider` when `OTEL_METRICS_EXPORTER=prometheus` is set;
  otherwise no listener starts, full stop.

## How to Test

- **Unit tests** — added `tests/node_agent_prometheus_exporter_test.yaml` (asserts the env
  var is present when enabled, absent by default). 
- Use helm template by adding `prometheusExporterEnabled` and verify if OTEL_METRICS_EXPORTER is present

- **End-to-end**, on a local Flux-managed kind cluster (kind v0.31.0, kubernetes v1.35.0,
Flux v2.7.5 — helm-controller v1.4.5, kustomize-controller v1.7.3, source-controller
v1.7.4 — Helm v4.1.0), with this branch wired in as a Flux `GitRepository` HelmRelease
source: deployed alongside kube-prometheus-stack chart 88.0.0 (v0.93.0) + Grafana chart
10.5.15 (v12.3.1), scraped node-agent's existing `ServiceMonitor`, and confirmed real
Prometheus-format output on `:8080/metrics`. 

## Examples/Screenshots

image

image

image

## Related issues/PRs:
Resolved #889 

put an [x] in the box to get it checked 

**Please open the PR against the `main` branch**

-->
 

## Summary by CodeRabbit

- **New Features**
  - Added Prometheus metrics exporter support for the node agent when enabled in configuration.

- **Bug Fixes**
  - Ensured the exporter environment setting is omitted when Prometheus export is disabled or left at its default.

- **Tests**
  - Added coverage verifying both enabled and default configurations.

### DIFF
--- a/charts/kubescape-operator/templates/node-agent/_node-agent.tpl
+++ b/charts/kubescape-operator/templates/node-agent/_node-agent.tpl
@@ -60,6 +60,10 @@ Parameters:
 - name: OTEL_COLLECTOR_SVC
   value: {{ .Values.configurations.otelUrl }}
 {{- end }}
+{{- if eq .Values.nodeAgent.config.prometheusExporter "enable" }}
+- name: OTEL_METRICS_EXPORTER
+  value: "prometheus"
+{{- end }}
 {{- if and .components.clamAV.enabled (not .autoscalerMode) }}
 - name: CLAMAV_SOCKET
   value: "/clamav/clamd.sock"
--- a/charts/kubescape-operator/tests/node_agent_prometheus_exporter_test.yaml
+++ b/charts/kubescape-operator/tests/node_agent_prometheus_exporter_test.yaml
@@ -37,3 +37,4 @@ tests:
           content:
             name: OTEL_METRICS_EXPORTER
             value: "prometheus"
+            
--- a/charts/kubescape-operator/tests/node_agent_prometheus_exporter_test.yaml
+++ b/charts/kubescape-operator/tests/node_agent_prometheus_exporter_test.yaml
@@ -1,0 +1,39 @@
+suite: node-agent Prometheus exporter OTEL env var
+templates:
+  - configs/components-configmap.yaml
+  - configs/cloudapi-configmap.yaml
+  - configs/cloud-secret.yaml
+  - configs/matchingRules-configmap.yaml
+  - node-agent/configmap.yaml
+  - node-agent/daemonset.yaml
+  - operator/configmap.yaml
+  - proxy-support/proxy-secret.yaml
+  - storage/certgen/configmap.yaml
+  - storage/configmap.yaml
+  - synchronizer/configmap.yaml
+  - operator/admission-webhook/configmap.yaml
+tests:
+  - it: sets OTEL_METRICS_EXPORTER when nodeAgent.config.prometheusExporter is enable
+    template: node-agent/daemonset.yaml
+    set:
+      unittest: true
+      clusterName: kind-kind
+      nodeAgent.config.prometheusExporter: enable
+    asserts:
+      - contains:
+          path: spec.template.spec.containers[0].env
+          content:
+            name: OTEL_METRICS_EXPORTER
+            value: "prometheus"
+
+  - it: does not set OTEL_METRICS_EXPORTER by default
+    template: node-agent/daemonset.yaml
+    set:
+      unittest: true
+      clusterName: kind-kind
+    asserts:
+      - notContains:
+          path: spec.template.spec.containers[0].env
+          content:
+            name: OTEL_METRICS_EXPORTER
+            value: "prometheus"


matthyx (commented):
Reviewed the change end-to-end (chart render + node-agent source + CI gates). The core fix is right: `nodeAgent.config.prometheusExporter` is the correct key (it already drives `prometheusExporterEnabled` in `node-agent/configmap.yaml`), putting it in the shared `node-agent.env` define means the daemonset, the multiplication daemonsets and the autoscaler `template-configmap.yaml` all get it, the `node-agent` Service already exposes port `prometheus`/8080 and `servicemonitor.yaml` already scrapes it, the node-agent netpol is egress-only so scraping isn't blocked, and the pinned image `v0.3.158` already contains the `OTEL_METRICS_EXPORTER == "prometheus"` branch in `pkg/otelsetup/setup.go`. `helm unittest charts/kubescape-operator` passes locally (39/39, snapshots unaffected).

**One blocker**: the new env var is not in the approved GKE Autopilot WorkloadAllowlist, so enabling the feature makes the node-agent unschedulable on Autopilot — and the `pr-allowlist-drift` gate is blind to it because `check-drift.sh` never enables this knob. Details inline, with the reproduction. The rest of the comments are non-blocking (docs on a silent OTLP-metrics trade-off, test scope).

charts/kubescape-operator/templates/node-agent/_node-agent.tpl
```diff
@@ -60,6 +60,10 @@ Parameters:
 - name: OTEL_COLLECTOR_SVC
   value: {{ .Values.configurations.otelUrl }}
 {{- end }}
+{{- if eq .Values.nodeAgent.config.prometheusExporter "enable" }}
+- name: OTEL_METRICS_EXPORTER
```
matthyx, line 64: **Blocker — GKE Autopilot allowlist drift.**

Autopilot admission is subset-based (`scripts/gke-allowlist/README.md`): a node-agent pod is admitted only if its env var *names* are a subset of the matched WorkloadAllowlist. `OTEL_METRICS_EXPORTER` is not listed for the `node-agent` container in `gke-allowlist/armo-kubescape-node-agent-1.40-v2.yaml` — the allowlist pinned by `nodeAgent.gke.allowlist.name`. So any Autopilot customer who sets `nodeAgent.config.prometheusExporter: enable` gets the node-agent DaemonSet rejected, i.e. the feature this PR enables is unusable there.

Reproduced on this branch (`f6d4d54`), same flags `check-drift.sh` uses plus the new knob:

```
$ python3 scripts/gke-allowlist/allowlist-drift.py \
    --chart charts/kubescape-operator \
    --allowlist gke-allowlist/armo-kubescape-node-agent-1.40-v2.yaml \
    --set server=api.example.com --set accessKey=... \
    --set capabilities.malwareDetection=enable --set capabilities.nodeSbomGeneration=enable \
    --set nodeAgent.sbomScanner.enabled=true --set configurations.otelUrl=http://otel-collector:4317 \
    --set global.httpsProxy=http://proxy:3128 --set global.overrideRuntimePath=/run/containerd/containerd.sock \
    --set nodeAgent.config.skipKernelVersionCheck=true --set nodeAgent.config.malwareScanAllFiles=true \
    --set nodeAgent.config.prometheusExporter=enable
CLASS: spec-drift  (NEW allowlist version required)
  - container 'node-agent': env OTEL_METRICS_EXPORTER not in allowlist
```

CI is green only because the gate doesn't exercise this path — `scripts/gke-allowlist/check-drift.sh` claims to render "the FULL set of optional features enabled", but it never sets `nodeAgent.config.prometheusExporter`. The unmodified check on this branch returns `CLASS: no-op` / rc=0.

Two things needed:

1. Get `- name: OTEL_METRICS_EXPORTER` into a new approved allowlist version for the `node-agent` container (Gerrit runbook in `scripts/gke-allowlist/README.md` § "When the drift gate fails"), then update the vendored copy under `gke-allowlist/` and the `nodeAgent.gke.allowlist.name` default in `values.yaml`.
2. Add `--set nodeAgent.config.prometheusExporter=enable` to `check-drift.sh` so this knob is covered by the gate from now on — otherwise the next env var added under the same `if` block repeats this silently.

charts/kubescape-operator/templates/node-agent/_node-agent.tpl
```diff
@@ -60,6 +60,10 @@ Parameters:
...
+  value: "prometheus"
```
matthyx, line 65: **Non-blocking — the trade-off here deserves a values.yaml comment.**

`pkg/otelsetup/setup.go` treats this as mutually exclusive with OTLP metric push: when `OTEL_METRICS_EXPORTER=prometheus` is set, "the OTLP metric exporter that go-logger may have configured is replaced by the prometheus reader". So a user who already has `configurations.otelUrl` set and flips `nodeAgent.config.prometheusExporter: enable` silently stops pushing metrics to their collector. Traces and logs are unaffected, which makes the loss easy to miss.

Second gap in the same area: the listener alone isn't enough to get data — `nodeAgent.serviceMonitor.enabled` defaults to `false`, so nothing scrapes `:8080/metrics` until the user also turns that on.

`nodeAgent.config.prometheusExporter` currently has no comment in `values.yaml` (unlike most neighbouring keys). Worth documenting both points there, and disambiguating it from `capabilities.prometheusExporter`, which is the separate prometheus-exporter deployment — easy to confuse.

FYI, not asking for a change: the `sbom-scanner` sidecar builds its own env list and doesn't get this var, so its metrics stay OTLP-only. That's the right call anyway (same pod netns, it could not bind `:8080` — `setup.go` even has a soft-fail path for exactly that), just noting it's a known gap in prometheus-only setups.

charts/kubescape-operator/tests/node_agent_prometheus_exporter_test.yaml
```diff
@@ -0,0 +1,40 @@
+suite: node-agent Prometheus exporter OTEL env var
+templates:
```
matthyx, line 2: **Non-blocking — this `templates:` list is copy-paste from another suite.**

Only `node-agent/daemonset.yaml` is ever rendered here; both tests override with `template:` anyway. The 11 other entries (`configs/*`, `operator/*`, `storage/*`, `proxy-support/*`, `synchronizer/*`) just make the suite slower to read and slower to run. Trim to:

```yaml
templates:
  - node-agent/daemonset.yaml
```

charts/kubescape-operator/tests/node_agent_prometheus_exporter_test.yaml
```diff
@@ -0,0 +1,40 @@
...
+            name: OTEL_METRICS_EXPORTER
+            value: "prometheus"
+
+  - it: does not set OTEL_METRICS_EXPORTER by default
+    template: node-agent/daemonset.yaml
+    set:
+      unittest: true
+      clusterName: kind-kind
+    asserts:
+      - notContains:
```
matthyx, line 35: **Non-blocking — the negative assertion is weaker than it looks, plus two coverage gaps.**

`notContains` matches the exact `name` + `value` pair, so it only fails if the var is emitted with precisely `value: "prometheus"`. A regression that emits `OTEL_METRICS_EXPORTER` with any other value passes this test. Assert absence by name only.

Also worth adding:
- an explicit `nodeAgent.config.prometheusExporter: disable` case — right now the negative test relies on the chart default, so it silently stops testing "disable" if the default ever flips;
- coverage of the autoscaler path. `node-agent/template-configmap.yaml` embeds the same `node-agent.env` block and is what the operator renders in autoscaler mode (`nodeAgent.autoscaler.enabled=true`); I confirmed the env var does land there, but nothing locks it in.

Nit: line 40 is still a whitespace-only line with no newline at EOF, despite f6d4d54 ("add newline at end of file") — `cat -A` shows the file ending in 12 spaces without a trailing `$`.

khuswant18:
Hey @matthyx, I was looking into #891 and noticed it’s been waiting on the author for a few weeks. If the original author isn’t continuing with it, would it be okay if I picked this up?

MihaiB-dev:
Hey, I started my studies so it is hard at the moment to work on this task, from my point of view, you can continue this. Thanks for helping!

khuswant18:
@MihaiB-dev No issues, will start working and raise PR soon

khuswant18:
@matthyx kindly close this PR

PR #893, opened 2026-08-03
feat(kubescape): add opt-in RBAC for agent runtime posture

dakshhhhh16:
## What this changes

Kubescape's in-cluster scanner needs permission to discover and read the custom resources that define agent sandbox posture. Without that access, a framework can work in CLI file scans while the operator silently misses the same resources in a live cluster.

This adds an opt-in `capabilities.agentRuntimePosture` chart value, set to `disable` by default. When set to `enable`, the Kubescape ClusterRole receives read-only (`get`, `list`, and `watch`) access to:

- `agents.x-k8s.io`: `sandboxes`
- `extensions.agents.x-k8s.io`: `sandboxtemplates`
- `ate.dev`: `actortemplates` and `workerpools`

These are the resources currently declared by Kubescape's scan contract. Keeping the capability disabled by default avoids expanding operator permissions before the agent runtime posture framework is available.

## Testing

- Added Helm unit coverage for both the default-disabled and explicitly-enabled configurations
- Ran the full chart unit suite: 47 tests and 996 snapshots passed
- Rendered the chart successfully with the feature enabled and disabled
- Ran `helm lint` successfully

## Summary by CodeRabbit

* **New Features**
  * Added an optional Agent Runtime posture capability for scanning Agent Sandbox and Agent Substrate resources.
  * The capability is disabled by default and can be enabled through Helm configuration.
  * When enabled, the operator receives read-only access to the required runtime resources.
* **Documentation**
  * Documented the new capability, its default setting, scan coverage, and configuration requirements.
* **Tests**
  * Added coverage confirming permissions remain disabled by default and are granted when enabled.

### DIFF
--- a/charts/kubescape-operator/templates/kubescape/clusterrole.yaml
+++ b/charts/kubescape-operator/templates/kubescape/clusterrole.yaml
@@ -86,17 +86,17 @@ rules:
 - apiGroups: ["kubescape.io"]
   resources: ["servicesscanresults"]
   verbs: ["get", "watch", "list"]
-{{- if .Values.agentRuntimePosture.enabled }}
+{{- if eq .Values.capabilities.agentRuntimePosture "enable" }}
 # Agent Sandbox and Agent Substrate resources are scanned for posture controls.
 # Keep these permissions opt-in until the agent runtime posture framework ships.
 - apiGroups: ["agents.x-k8s.io"]
   resources: ["sandboxes"]
   verbs: ["get", "list", "watch"]
 - apiGroups: ["extensions.agents.x-k8s.io"]
-  resources: ["sandboxclaims", "sandboxtemplates", "sandboxwarmpools"]
+  resources: ["sandboxtemplates"]
   verbs: ["get", "list", "watch"]
 - apiGroups: ["ate.dev"]
-  resources: ["actortemplates", "workerpools", "sandboxconfigs"]
+  resources: ["actortemplates", "workerpools"]
   verbs: ["get", "list", "watch"]
 {{- end }}
 {{- if eq .Values.capabilities.riskAcceptance "enable" }}
--- a/charts/kubescape-operator/templates/kubescape/clusterrole.yaml
+++ b/charts/kubescape-operator/templates/kubescape/clusterrole.yaml
@@ -86,6 +86,19 @@ rules:
 - apiGroups: ["kubescape.io"]
   resources: ["servicesscanresults"]
   verbs: ["get", "watch", "list"]
+{{- if .Values.agentRuntimePosture.enabled }}
+# Agent Sandbox and Agent Substrate resources are scanned for posture controls.
+# Keep these permissions opt-in until the agent runtime posture framework ships.
+- apiGroups: ["agents.x-k8s.io"]
+  resources: ["sandboxes"]
+  verbs: ["get", "list", "watch"]
+- apiGroups: ["extensions.agents.x-k8s.io"]
+  resources: ["sandboxclaims", "sandboxtemplates", "sandboxwarmpools"]
+  verbs: ["get", "list", "watch"]
+- apiGroups: ["ate.dev"]
+  resources: ["actortemplates", "workerpools", "sandboxconfigs"]
+  verbs: ["get", "list", "watch"]
+{{- end }}
 {{- if eq .Values.capabilities.riskAcceptance "enable" }}
 # SecurityException CRDs (opt-in via capabilities.riskAcceptance). The in-cluster
 # posture scanner reads these to suppress matching findings, and emits Events on the
--- a/charts/kubescape-operator/values.yaml
+++ b/charts/kubescape-operator/values.yaml
@@ -23,11 +23,6 @@ clusterName: "cluster"
 # Note: when operator.triggerSecurityFramework is true (default), "security" is also scanned.
 defaultFrameworks: []
 
-# Agent runtime posture support
-agentRuntimePosture:
-  # -- Grant the Kubescape scanner read-only access to Agent Sandbox and Agent Substrate CRDs
-  enabled: false
-
 # The namespace for running the chart
 # If you wish to run the chart in a different namespace, make sure to update this value
 ksNamespace: kubescape
@@ -147,6 +142,14 @@ capabilities:
   seccompProfileBackend: crd
   manageWorkloads: disable
   syncSBOM: disable
+
+  # ====== Agent runtime posture ======
+  # Grants the scanner read-only access to the Agent Sandbox and Agent Substrate
+  # CRDs currently covered by the scan contract. Tracking issue:
+  # https://github.com/kubescape/kubescape/issues/2557
+  # Keep disabled until the Agent Runtime Hardening framework ships and is enabled
+  # by default for operator scans; that rollout is what should flip this default.
+  agentRuntimePosture: disable
 
   # ====== Risk acceptance ======
   riskAcceptance: disable
--- a/charts/kubescape-operator/values.yaml
+++ b/charts/kubescape-operator/values.yaml
@@ -23,6 +23,11 @@ clusterName: "cluster"
 # Note: when operator.triggerSecurityFramework is true (default), "security" is also scanned.
 defaultFrameworks: []
 
+# Agent runtime posture support
+agentRuntimePosture:
+  # -- Grant the Kubescape scanner read-only access to Agent Sandbox and Agent Substrate CRDs
+  enabled: false
+
 # The namespace for running the chart
 # If you wish to run the chart in a different namespace, make sure to update this value
 ksNamespace: kubescape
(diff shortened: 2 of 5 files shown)


dakshhhhh16:
@matthyx 
You can have a look into this, it would be useful here ig, I can remove the comments if not required

matthyx (commented):
Thanks for this — the API groups and resource names all check out, and I verified the chart side locally on your branch: `helm unittest` is green (47 tests / 996 snapshots), `helm lint` is clean, rendering with `agentRuntimePosture.enabled=true` emits exactly the three rules, and the default render emits none.

I also cross-checked the groups against upstream so we don't ship a typo'd ClusterRole:

- `kubernetes-sigs/agent-sandbox` → `k8s/crds/agents.x-k8s.io_sandboxes.yaml`, `extensions.agents.x-k8s.io_sandboxclaims.yaml`, `..._sandboxtemplates.yaml`, `..._sandboxwarmpools.yaml`
- `agent-substrate/substrate` → `manifests/ate-install/generated/ate.dev_actortemplates.yaml`, `ate.dev_sandboxconfigs.yaml`, `ate.dev_workerpools.yaml`

Every group/resource pairing in the diff is correct. Two blockers before this can go in, then some smaller points.

## Blockers

**1. The values shape doesn't follow the chart's convention.** Every other opt-in toggle that gates RBAC in this chart lives under `capabilities:` as an `enable`/`disable` string — including the closest analogue, `capabilities.riskAcceptance`, which gates the SecurityException rules immediately below your new block in the same file. A top-level `agentRuntimePosture.enabled` bool adds a second toggle idiom for the same kind of thing. Values are public API, so renaming it after a release is a breaking change for users — better to get it right now. Please move it to `capabilities.agentRuntimePosture: disable` and gate with `{{- if eq .Values.capabilities.agentRuntimePosture "enable" }}`.

**2. The PR is still a draft.** CodeRabbit skipped its pass ("Draft detected") and it can't be merged in this state. Mark it ready for review once the above is addressed, so the automated review runs on it too.

## Should fix

**3. The grant is wider than what the scanner actually declares.** The scan contract in `kubescape/kubescape` (`core/pkg/resourcehandler/crd_scan_contract_test.go`, `agentRuntimeMatches()`) covers only `agents.x-k8s.io` Sandbox, `extensions.agents.x-k8s.io` SandboxTemplate, and `ate.dev` ActorTemplate + WorkerPool. This PR additionally grants `sandboxclaims`, `sandboxwarmpools` and `sandboxconfigs`. Either drop those three — least privilege, and we can widen when a control needs them — or state in the comment that they're deliberately pre-granted for upcoming controls.

**4. Nothing consumes the flag yet.** There is no agent-runtime framework in `kubescape/regolibrary` today, so enabling this changes permissions and nothing else. That matches your stated intent, but please link the tracking issue or the framework PR from the values comment, and note what flips the default later — otherwise this becomes orphaned config that nobody dares remove.

## Answers and nits

- **Keep the comments in `clusterrole.yaml`** — you asked whether to remove them. Don't; that block mirrors what `capabilities.riskAcceptance` does right below it, and it's the only place a reader learns why those rules are conditional.
- Optional, not for this PR: the repo keeps short feature pages under `docs/features/`. One for agent runtime posture would fit nicely once the framework actually lands.

Test coverage and README entry are both fine as-is — the enabled/disabled pair is exactly the right shape, and it'll carry over unchanged apart from the key rename.

charts/kubescape-operator/values.yaml
```diff
@@ -23,6 +23,11 @@ clusterName: "cluster"
 # Note: when operator.triggerSecurityFramework is true (default), "security" is also scanned.
 defaultFrameworks: []
 
+# Agent runtime posture support
+agentRuntimePosture:
```
matthyx: Blocker: please put this under the existing `capabilities:` block instead of a new top-level key, as an enable/disable string:

```yaml
capabilities:
  # ...
  # ====== Agent runtime posture ======
  # Grants the scanner read-only access to Agent Sandbox and Agent Substrate CRDs.
  agentRuntimePosture: disable
```

`capabilities.riskAcceptance` is the exact precedent — an opt-in string toggle whose only effect is extra read-only RBAC in `kubescape/clusterrole.yaml`. Values are public API, so this is much cheaper to change now than after a release.

charts/kubescape-operator/templates/kubescape/clusterrole.yaml
```diff
@@ -86,6 +86,19 @@ rules:
 - apiGroups: ["kubescape.io"]
   resources: ["servicesscanresults"]
   verbs: ["get", "watch", "list"]
+{{- if .Values.agentRuntimePosture.enabled }}
```
matthyx: Follows from the values comment: this becomes `{{- if eq .Values.capabilities.agentRuntimePosture "enable" }}`, which also makes it read consistently with the `riskAcceptance` gate 13 lines below.

charts/kubescape-operator/templates/kubescape/clusterrole.yaml
```diff
@@ -86,6 +86,19 @@ rules:
...
+# Agent Sandbox and Agent Substrate resources are scanned for posture controls.
+# Keep these permissions opt-in until the agent runtime posture framework ships.
+- apiGroups: ["agents.x-k8s.io"]
+  resources: ["sandboxes"]
+  verbs: ["get", "list", "watch"]
+- apiGroups: ["extensions.agents.x-k8s.io"]
+  resources: ["sandboxclaims", "sandboxtemplates", "sandboxwarmpools"]
```
matthyx: `sandboxclaims` and `sandboxwarmpools` are real CRDs in that group, but the scanner does not ask for them: `agentRuntimeMatches()` in `kubescape/kubescape` `core/pkg/resourcehandler/crd_scan_contract_test.go` declares only `SandboxTemplate` for `extensions.agents.x-k8s.io`. Suggest narrowing to `["sandboxtemplates"]` for now, or noting in the comment that the other two are pre-granted on purpose.

charts/kubescape-operator/templates/kubescape/clusterrole.yaml
```diff
@@ -86,6 +86,19 @@ rules:
...
+  verbs: ["get", "list", "watch"]
+- apiGroups: ["ate.dev"]
+  resources: ["actortemplates", "workerpools", "sandboxconfigs"]
```
matthyx: Same point here: the contract declares `ActorTemplate` and `WorkerPool` for `ate.dev`, not `SandboxConfig`. Either drop `sandboxconfigs` or say why it is included ahead of the controls that read it.

matthyx (approved):
All four points addressed cleanly in `f986cab` — thanks for the quick turnaround.

- **Values shape** — now `capabilities.agentRuntimePosture: disable`, gated with `{{- if eq .Values.capabilities.agentRuntimePosture "enable" }}`, exactly matching the `riskAcceptance` idiom right below it. README row updated to the new key with the right type/default, and the `ks-capabilities` snapshots regenerated for the new entry.
- **Scope** — narrowed to `sandboxes`, `sandboxtemplates`, `actortemplates`, `workerpools`, which is now a precise match for `agentRuntimeMatches()` in the scan contract. Nothing granted that the scanner doesn't ask for.
- **Traceability** — kubescape/kubescape#2557 linked from the values comment, with a note that the Agent Runtime Hardening framework rollout is what flips the default. That's the right owner for the flag.
- **Draft** — lifted, CodeRabbit is running now.

Re-verified locally on the head commit: `helm unittest` 47 tests / 996 snapshots pass, `helm lint` clean, `--set capabilities.agentRuntimePosture=enable` renders exactly the three rules and the default render emits none. The snapshot diff is only the new capabilities-map entry plus the resulting `checksum/capabilities-config` churn — expected, and the same one-time rolling restart on upgrade that any capabilities addition causes.

LGTM. Approving.

dakshhhhh16:
> All four points addressed cleanly in `f986cab` — thanks for the quick turnaround.
> 
> * **Values shape** — now `capabilities.agentRuntimePosture: disable`, gated with `{{- if eq .Values.capabilities.agentRuntimePosture "enable" }}`, exactly matching the `riskAcceptance` idiom right below it. README row updated to the new key with the right type/default, and the `ks-capabilities` snapshots regenerated for the new entry.
> * **Scope** — narrowed to `sandboxes`, `sandboxtemplates`, `actortemplates`, `workerpools`, which is now a precise match for `agentRuntimeMatches()` in the scan contract. Nothing granted that the scanner doesn't ask for.
> * **Traceability** — [kubescape/kubescape#2557](https://github.com/kubescape/kubescape/issues/2557) linked from the values comment, with a note that the Agent Runtime Hardening framework rollout is what flips the default. That's the right owner for the flag.
> * **Draft** — lifted, CodeRabbit is running now.
> 
> Re-verified locally on the head commit: `helm unittest` 47 tests / 996 snapshots pass, `helm lint` clean, `--set capabilities.agentRuntimePosture=enable` renders exactly the three rules and the default render emits none. The snapshot diff is only the new capabilities-map entry plus the resulting `checksum/capabilities-config` churn — expected, and the same one-time rolling restart on upgrade that any capabilities addition causes.
> 
> LGTM. Approving.

Letsgoo!

PR #894, opened 2026-08-03
feat(storage): allow running the APIServer on the host network

### LINKED ISSUE #453: Allow hostNetwork for storage deployment
## Overview
Currently the helm chart does not provide the ability to set hostNetwork to true in the storage deployment. Also the port defaults to 443, which might already be in use if hostNetwork is set to true.

## Problem
When using a virtual Network with custom CNIs like Cilium the storage API endpoint is not reachable by the kubeapi server. 

## Solution
Template values to set hostNetwork and port in the helm chart

## Alternatives
None

## Additional context
None

ahauserv:
## Overview
Adds `storage.hostNetwork` (default `false`) so the storage APIServer can run on the host
network. Without it, the aggregated APIService stays unavailable on clusters whose control
plane has no route to pod addresses, and no posture result is ever written or read.

Nothing signals the failure: every component reports `Running` and the PVC binds. The API
group simply isn't served, so:

* `kubectl get workloadconfigurationscansummaries` returns `the server doesn't have a resource type`
* `prometheus-exporter` logs a failure on every poll
* on EKS the apiserver reports `Available: False … Address is not allowed`

Enabling it also sets `dnsPolicy: ClusterFirstWithHostNet`; without that the pod inherits the
node's `resolv.conf` and can no longer resolve cluster Services. Note that
`storage.serverPort` must then be free on every node the component can schedule to.

## How to Test
```
helm template test charts/kubescape-operator --kube-version 1.28.0 --set clusterName=x | grep -E 'hostNetwork: |dnsPolicy: '
helm template test charts/kubescape-operator --kube-version 1.28.0 --set clusterName=x --set storage.hostNetwork=true | grep -E 'hostNetwork: |dnsPolicy: '
```
The first command prints nothing. The second prints `hostNetwork: true` and
`dnsPolicy: ClusterFirstWithHostNet` on the storage deployment.

## Related issues/PRs:

* Resolves #453
* Prior art: [cert-manager#3113](https://github.com/cert-manager/cert-manager/pull/3113) added
  the same option for its webhook, and
  [cert-manager#6156](https://github.com/cert-manager/cert-manager/pull/6156) added the
  `ClusterFirstWithHostNet` pairing in a follow-up, so this PR includes it from the start.
  [aws/eks-charts#336](https://github.com/aws/eks-charts/pull/336) documents the same condition
  for a custom CNI on EKS.

## Summary by CodeRabbit

* **New Features**
  * Added an optional `storage.hostNetwork` setting for deploying the storage API server on the host network.
  * Configured compatible DNS behavior when host networking is enabled.
  * The option is disabled by default and can support environments where control-plane access to pod IPs is limited.

* **Documentation**
  * Documented the new storage networking option and its configuration details.

* **Tests**
  * Added coverage for default behavior, host-network activation, DNS settings, and probe port preservation.

### DIFF
--- a/charts/kubescape-operator/values.yaml
+++ b/charts/kubescape-operator/values.yaml
@@ -539,6 +539,8 @@ storage:
       retries: 60
       retryIntervalSeconds: 5
   serverPort: 8443
+  # -- Bind the storage APIServer to the host network. Required when using a custom CNI where the control plane cannot reach pod IPs
+  hostNetwork: false
   service:
     type: ClusterIP
     annotations: {}
--- a/charts/kubescape-operator/README.md
+++ b/charts/kubescape-operator/README.md
@@ -162,6 +162,7 @@ However, we recommend that you give Kubescape no less than 500m CPU no matter th
 | operator.nodeSelector | object | `{}` | [Node selector](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/) |
 | operator.volumes | object | `[]` | Additional volumes for the web socket |
 | operator.volumeMounts | object | `[]` | Additional volumeMounts for the web socket |
+| storage.hostNetwork | bool | `false` | Bind the storage APIServer to the host network. Required when using a custom CNI where the control plane cannot reach pod IPs |
 | nodeAgent.autoscaler.bottlerocketAutoDetect | bool | `true` | When the autoscaler is enabled, auto-detect AWS Bottlerocket nodes and set `seLinuxType: super_t` on the node-agent DaemonSet rendered for that node group, so you don't need to `--set nodeAgent.seLinuxType=super_t` manually |
 | prometheusExporter.serviceMonitor.enabled | bool | `false` | enable/disable service monitor for prometheus-exporter integration |
 | awsIamRoleArn | string | `nil` | AWS IAM arn role |
--- a/charts/kubescape-operator/templates/storage/deployment.yaml
+++ b/charts/kubescape-operator/templates/storage/deployment.yaml
@@ -48,6 +48,10 @@ spec:
       {{- end }}
       {{- end }}
       serviceAccountName: {{ .Values.storage.name }}
+      {{- if .Values.storage.hostNetwork }}
+      hostNetwork: true
+      dnsPolicy: ClusterFirstWithHostNet
+      {{- end }}
       securityContext:
         seccompProfile:
           type: RuntimeDefault
--- a/charts/kubescape-operator/tests/storage_host_network_test.yaml
+++ b/charts/kubescape-operator/tests/storage_host_network_test.yaml
@@ -1,0 +1,51 @@
+suite: Storage hostNetwork tests
+tests:
+  - it: keeps the storage APIServer off the host network by default
+    template: storage/deployment.yaml
+    capabilities:
+      apiVersions:
+        - batch/v1
+    set:
+      unittest: true
+      clusterName: kind-kind
+    asserts:
+      - notExists:
+          path: spec.template.spec.hostNetwork
+      - notExists:
+          path: spec.template.spec.dnsPolicy
+
+  - it: runs the storage APIServer on the host network when enabled
+    template: storage/deployment.yaml
+    capabilities:
+      apiVersions:
+        - batch/v1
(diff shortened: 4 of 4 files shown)


matthyx:
@ahauserv are you sure it's the issue? does it solve it? I know our current storage doesn't run on 1.28 but I think it's because of incompatibilities with the api-server

ahauserv:
@matthyx I tested it on EKS 1.35 and it failed without this commit. With this commit the `Available: False … Address is not allowed` is gone and I get `HTTP 200`
I added `--kube-version 1.28.0` to the test command because the Chart.yaml said `requires Kubernetes >= 1.28`

charts/kubescape-operator/values.yaml
```diff
@@ -539,6 +539,8 @@ storage:
       retries: 60
       retryIntervalSeconds: 5
   serverPort: 8443
+  # -- Bind the storage APIServer to the host network. Required when using a custom CNI where the control plane cannot reach pod IPs
+  hostNetwork: false
```
matthyx, line 543: Please fold two constraints into this description — this comment and the README row are the only place a user enabling the flag will look.

1. `storage.serverPort` (8443) must be free on **every node the pod can schedule to**. This is already in the PR description; it belongs here. (Same point CodeRabbit raised.)

2. Enabling this makes `templates/storage/networkpolicy.yaml` a no-op. That policy selects the storage pod with a `podSelector` and restricts ingress on `storage.serverPort` to `global.networkPolicy.apiServerIP`. NetworkPolicy is enforced on the pod's own interface, and a host-network pod doesn't have one — Calico won't apply it without host endpoints, Cilium won't without the host firewall. So with `storage.hostNetwork=true` **and** `global.networkPolicy.enabled=true`, 8443 ends up listening on every node IP with only the node's firewall in front of it.

The default `storage.mtls.enabled=true` largely covers (2), since `configmap.yaml` then sets `tlsClientCaFile` and the server requires a client cert. The combination that actually worries me is `hostNetwork=true` + `mtls.enabled=false`. A sentence here is enough; I don't think the chart needs to block it.

matthyx (approved):
Approving — the change is opt-in, defaults to off, and is a no-op for existing users. Nothing here blocks the merge; the notes below are documentation, not code.

### What I checked out and verified locally

- Rendering matches what you describe: default output contains neither `hostNetwork` nor `dnsPolicy`; `--set storage.hostNetwork=true` emits both at the pod-spec level.
- `helm unittest charts/kubescape-operator/` — 48 tests and 996 snapshots pass. Snapshots are unchanged, which is the expected consequence of defaulting to `false`.
- No other component in the chart sets `hostNetwork`, so there's no existing convention this diverges from.
- The storage Deployment is `replicas: 1` with `strategy: Recreate`, so a rollout never has two pods alive at once and can't collide with its own host port. Worth stating explicitly since that's the usual failure mode for host-networked Deployments.
- Pairing `ClusterFirstWithHostNet` is correct and matches the cert-manager precedent you linked; the probes stay on `storage.serverPort`, and the port is >1024 so `runAsUser: 65532` can still bind it.

### Non-blocking

1. **Document the two constraints in `values.yaml` / README** — host-port availability on every eligible node, and the fact that the storage NetworkPolicy stops being enforced. Details in the inline comment on `values.yaml`.
2. **README row placement** — `charts/kubescape-operator/README.md:165` lands in the middle of the `operator.*` block. This README isn't helm-docs generated, so ordering is manual; please move it out of that group.
3. **CodeRabbit's test nit** (assert `serverBindPort` in the ConfigMap and `targetPort` in the Service) — your call. Both are driven by the same `storage.serverPort` and are pre-existing behavior this PR doesn't touch, so I'd skip it rather than grow the test.

### CI

`pull_request_created` and `pr-allowlist-drift` are both stuck at `action_required` — the helm unittest job has never actually run on this branch. I'm approving the workflow runs. The diff touches no workflow files or scripts, and I've already run the same suite locally.

### On my earlier question

Your EKS 1.35 reproduction answers it. I was thinking of a separate storage/apiserver compatibility problem; this is an independent, opt-in escape hatch for the control-plane-can't-reach-pod-IPs case, so it stands on its own either way. Thanks for the writeup and the prior art.

PR #850, opened 2026-05-29
prepate release 1.40.3

matthyx:
## Summary by CodeRabbit

* **Chores**
  * Released Kubescape Operator v1.40.3.
  * Updated container image tags for kubescape, operator, kubevuln, storage, node-agent, synchronizer, prometheus-exporter, and scheduler helper.
  * Updated Helm documentation and version badges to v1.40.3.
  * Adjusted scan capability enablement and storage behavior to be driven directly by chart values.

PR #896, opened 2026-08-10
fix(components): re-gate node/config/vuln scans and storage on backend-storage

AlonLiwsky:
## Overview

**Current behavior.** With `capabilities.backend-storage=enable`, the chart deploys
`kubevuln`, `kubevuln-scheduler`, `kubescape`, `kubescape-scheduler`, the in-cluster
`storage` aggregated APIServer and its `APIService` — while the *same* render emits:

- `effectiveCapabilities.vulnerabilityScan: backend` (and `configurationScan`, `nodeScan`)
  into `ks-capabilities`
- `backendStorageEnabled: true` into the node-agent `config.json`

So the cluster is told scanning happens in the backend, and then also runs the in-cluster
scanners. node-agent ships SBOMs to the backend while a local `kubevuln` scans the same
workloads — double scanning, plus a storage server and a cluster-scoped `APIService` that
the mode is meant to remove.

**Future behavior.** The three scan flags are ANDed with `backendStorageEnabled` again and
`storage.enabled` is gated on it, so the rendered workloads match the
`effectiveCapabilities` contract the chart already emits.

This restores gating removed in #850 (`79bd478f`) while **keeping** the explicit
`storage.enabled` knob that PR introduced — the knob still works for everyone not using
backend-storage.

```diff
-{{- $nodeScanEnabled := eq .Values.capabilities.nodeScan "enable" }}
-{{- $configurationScanEnabled := eq .Values.capabilities.configurationScan "enable" }}
-{{- $vulnerabilityScanEnabled := eq .Values.capabilities.vulnerabilityScan "enable" }}
+{{- $nodeScanEnabled := and (eq .Values.capabilities.nodeScan "enable") (not $configurations.backendStorageEnabled) }}
+{{- $configurationScanEnabled := and (eq .Values.capabilities.configurationScan "enable") (not $configurations.backendStorageEnabled) }}
+{{- $vulnerabilityScanEnabled := and (eq .Values.capabilities.vulnerabilityScan "enable") (not $configurations.backendStorageEnabled) }}
 storage:
-  enabled: {{ .Values.storage.enabled }}
+  enabled: {{ and .Values.storage.enabled (not $configurations.backendStorageEnabled) }}
```

## Additional Information

`$nodeScanEnabled` is assigned and never referenced — in `1.40.2` and at HEAD both (host
scanner removed in #773). It is re-gated only for consistency with the other two; it
renders nothing either way.

`capabilities.gates` (#852, `a94aabc7`, 2026-06-05) predates the decouple (#850,
2026-07-22) by seven weeks, and still emits `...Scan: backend` under backend-storage. Its
own comment says it exists so "the rendered value and the warning about that value can
never drift apart. See issue #851" — and #851 is the chart's `capabilities` block
disagreeing with the rendered config about whether a feature is on. Without this patch,
1.40.3 reintroduces that drift for `vulnerabilityScan` and `configurationScan`.

No value combination is a workaround: `enable` gives the right signal and the wrong
workloads; `disable` + `storage.enabled=false` gives the right workloads and the wrong
signal; and downstream wrappers can't express "off only when backend-storage is on"
in values at all.

## How to Test

```bash
helm template t charts/kubescape-operator \
  --set account=<guid> --set accessKey=<key> \
  --set server=api.armosec.io --set clusterName=c \
  --set capabilities.backend-storage=enable \
  | grep -E '^kind:|^  name:'
```

Verified against `6a0fb27`:

| case | result |
|---|---|
| no `backend-storage` (default install) | **identical render** (modulo per-render `genSelfSignedCert`/`genCA`) |
| Rapid7 value set (`backend-storage=enable`, scans `disable`, `storage.enabled=false`) | **identical render — strict no-op** |
| `backend-storage=enable`, minimal values | 75 → 48 documents |
| snapshot-test value set | 9 → 3 real workloads (`node-agent`, `operator`, `synchronizer`) |
| `kubescape-operator-1.40.2`, same values | same 3 workloads — behavior restored |
| `helm unittest` | 53/53 pass; 1 of 22 snapshot entries changes |

The snapshot case is renamed back to "backend-storage enabled **disables** scanning
capabilities", reverting the rename in #850.

Note: `helm unittest` v1.0.0 prunes the vanished snapshot documents and still reports
PASS, so the regenerated `snapshot_test.yaml.snap` is committed here deliberately rather
than left to drift.

## Release

No `Chart.yaml` bump here, to stay out of the way of the usual "prepare release" flow.
@matthyx — could this go out as `1.40.4`? `1.40.3` is the latest release and HEAD is still
versioned `1.40.3`, so there's currently no released version carrying the fix, and
downstream wrappers pin the dependency exactly.

## Related issues/PRs

* Regression introduced in #850
* Same drift class as #851

### DIFF
--- a/charts/kubescape-operator/templates/_common.tpl
+++ b/charts/kubescape-operator/templates/_common.tpl
@@ -63,9 +63,9 @@ submit: {{ $submit }}
 
 {{- define "components" -}}
 {{- $configurations := fromYaml (include "configurations" .) }}
-{{- $nodeScanEnabled := eq .Values.capabilities.nodeScan "enable" }}
-{{- $configurationScanEnabled := eq .Values.capabilities.configurationScan "enable" }}
-{{- $vulnerabilityScanEnabled := eq .Values.capabilities.vulnerabilityScan "enable" }}
+{{- $nodeScanEnabled := and (eq .Values.capabilities.nodeScan "enable") (not $configurations.backendStorageEnabled) }}
+{{- $configurationScanEnabled := and (eq .Values.capabilities.configurationScan "enable") (not $configurations.backendStorageEnabled) }}
+{{- $vulnerabilityScanEnabled := and (eq .Values.capabilities.vulnerabilityScan "enable") (not $configurations.backendStorageEnabled) }}
 kubescape:
   enabled: {{ $configurationScanEnabled }}
 kubescapeScheduler:
@@ -89,7 +89,7 @@ operator:
 serviceDiscovery:
   enabled: {{ $configurations.submit }}
 storage:
-  enabled: {{ .Values.storage.enabled }}
+  enabled: {{ and .Values.storage.enabled (not $configurations.backendStorageEnabled) }}
 prometheusExporter:
   enabled: {{ eq .Values.capabilities.prometheusExporter "enable" }}
 cloudSecret:
--- a/charts/kubescape-operator/tests/__snapshot__/snapshot_test.yaml.snap
+++ b/charts/kubescape-operator/tests/__snapshot__/snapshot_test.yaml.snap
@@ -15748,20 +15748,12 @@ autoscaler mode without sbom sidecar:
         tier: ks-control-plane
       name: synchronizer
       namespace: kubescape
-backend-storage enabled allows scanning capabilities:
+backend-storage enabled disables scanning capabilities:
   1: |+
     raw: |+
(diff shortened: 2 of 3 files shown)


matthyx (dismissed):
@AlonLiwsky we need to discuss about what we want here, because we added the requirement on storage.disable=true for a reason

PR #897, opened 2026-08-11
fix: drop all Linux capabilities in workload container securityContexts

ahauserv:
## Overview

Adds `capabilities: drop [ALL]` to the 12 workload container securityContexts.
Every container already sets `allowPrivilegeEscalation: false`, `runAsNonRoot: true` with a `RuntimeDefault` seccomp profile and none of the components uses a capability.

The missing drop is the one field that keeps the chart out of namespaces with `pod-security.kubernetes.io/enforce: restricted`.
Admission rejects every pod with:

```
violates PodSecurity "restricted:latest": unrestricted capabilities
(container "kubescape" must set securityContext.capabilities.drop=["ALL"])
```

It is also the finding Kubescape raises against its own namespace when it scans the cluster it runs in.

The certgen init containers already declare the drop https://github.com/kubescape/helm-charts/blob/main/charts/kubescape-operator/templates/storage/deployment.yaml#L72 and is untouched.

## How to Test

New tests in `tests/container_capabilities_test.yaml` assert the drop on 5 out of 12 deployed components (if needed I can add all 12). The 104 snapshot changes are exactly the capability additions.

Confirmed in a cluster: all five deployed components run with the drop, scans complete and persist their results.

## Related issues/PRs:

* Related: #740 asked for configurable securityContexts. The hardcoded contexts were already nearly restricted-complete, so finishing them fixes the restricted-namespace case without adding a knob.

## Summary by CodeRabbit

* **Security Enhancements**
  * Containers across Kubescape Operator components now drop all Linux capabilities by default.
  * Applies to scheduled jobs, services, database components, scanners, exporters, and supporting workloads.
  * The autoupdater uses the `RuntimeDefault` seccomp profile.
  * Reduces the potential impact of container compromise through stricter runtime permissions.

* **Tests**
  * Added coverage to verify that required containers run without Linux capabilities.

### DIFF
--- a/charts/kubescape-operator/templates/autoupdater/cronjob.yaml
+++ b/charts/kubescape-operator/templates/autoupdater/cronjob.yaml
@@ -30,6 +30,8 @@ spec:
             runAsUser: 1000
             runAsGroup: 1000
             fsGroup: 1000
+            seccompProfile:
+              type: RuntimeDefault
           volumes:
             - name: "helm-scratch-data"
               emptyDir:
--- a/charts/kubescape-operator/templates/autoupdater/cronjob.yaml
+++ b/charts/kubescape-operator/templates/autoupdater/cronjob.yaml
@@ -41,6 +41,9 @@ spec:
             securityContext:
               allowPrivilegeEscalation: false
               readOnlyRootFilesystem: true
+              capabilities:
+                drop:
+                  - ALL
             volumeMounts:
               - name: "helm-scratch-data"
                 mountPath: /data/helm-scratch-data
--- a/charts/kubescape-operator/templates/grype-offline-db/cronjob.yaml
+++ b/charts/kubescape-operator/templates/grype-offline-db/cronjob.yaml
@@ -54,6 +54,9 @@ spec:
               readOnlyRootFilesystem: true
               runAsNonRoot: true
               runAsUser: 100
+              capabilities:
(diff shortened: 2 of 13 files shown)


ahauserv:
@coderabbitai review

matthyx (approved):
Reviewed the diff, the earlier CodeRabbit thread, and CI status.

- The `capabilities.drop: [ALL]` additions are consistent with the existing `allowPrivilegeEscalation: false` / `runAsNonRoot: true` contexts already in place on every touched container, so this closes the gap for `restricted` PodSecurity namespaces without changing runtime behavior.
- The earlier blocking finding (missing `seccompProfile.type: RuntimeDefault` on the autoupdater CronJob pod spec) was already fixed in a follow-up commit and the thread is resolved.
- All required checks (DCO, GitGuardian, allowlist-drift, FOSSA, CodeRabbit) are green, and the snapshot diff only reflects the expected capability additions.

Non-blocking nitpick (same one CodeRabbit raised): `tests/container_capabilities_test.yaml` only asserts the drop on 5 of the 12 touched containers (synchronizer, kubevuln + sbom-scanner, both Grype resources, autoupdater, kubevuln-scheduler are untested). Worth adding in a follow-up so a future regression on those templates would actually fail CI, but it's not a reason to hold this PR.

LGTM, approving.

PR #899, opened 2026-08-13
fix: forward HTTPS_PROXY/no_proxy to the sbom-scanner sidecar

matthyx:
## Overview
Follow-up to #898. The `sbom-scanner` sidecar (added to `node-agent` and `kubevuln` for OOM isolation) pulls image data over HTTPS itself but never received `global.httpsProxy`/`no_proxy` the way the main container does. In clusters that route egress through a proxy — especially with `global.networkPolicy.createEgressRules` enabled, which only opens egress to DNS, the API server, and the proxy IP/port — the sidecar had no route to the registry at all.

Adds the same `HTTPS_PROXY`/`no_proxy` env vars to the `sbom-scanner` container in:
- `charts/kubescape-operator/templates/kubevuln/deployment.yaml`
- `charts/kubescape-operator/templates/node-agent/_node-agent.tpl`

## How to Test
```
helm unittest charts/kubescape-operator
```
Added two new test cases in `tests/snapshot_test.yaml` asserting the sidecar picks up `HTTPS_PROXY`/`no_proxy` when `global.httpsProxy` is set. Full suite (60 tests / 996 snapshots) passes unchanged plus the 2 new tests.

AI-skills: none

## Summary by CodeRabbit

* **New Features**
  * SBOM scanner components now support configured HTTPS proxy settings.
  * Added support for combined proxy bypass rules through the `no_proxy` environment variable.
  * Proxy settings are applied conditionally when an HTTPS proxy is configured.

* **Tests**
  * Added coverage verifying proxy environment variables for SBOM scanner components.

### DIFF
--- a/charts/kubescape-operator/templates/kubevuln/deployment.yaml
+++ b/charts/kubescape-operator/templates/kubevuln/deployment.yaml
@@ -178,6 +178,12 @@ spec:
               value: "{{ .Values.logger.level }}"
             - name: KS_LOGGER_NAME
               value: "sbom-scanner"
+            {{- if ne .Values.global.httpsProxy "" }}
+            - name: HTTPS_PROXY
+              value: "{{ .Values.global.httpsProxy }}"
+            - name : no_proxy
+              value: "{{ $no_proxy_envar_list }}"
+            {{- end }}
           volumeMounts:
             - name: sbom-comm
               mountPath: /sbom-comm
--- a/charts/kubescape-operator/templates/node-agent/_node-agent.tpl
+++ b/charts/kubescape-operator/templates/node-agent/_node-agent.tpl
@@ -266,6 +266,7 @@ SBOM Scanner Sidecar Container (optional)
 Parameters:
   - Values: .Values
   - components: $components
+  - no_proxy_envar_list: the no_proxy list
 */}}
 {{- define "node-agent.sbomScannerContainer" -}}
 {{- if .components.sbomScanner.enabled }}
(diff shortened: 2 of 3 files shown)


matthyx:
cc @slashben we'll need a new autopilot submission

matthyx (commented):
Code review: this looks correct and well-tested.

- `node-agent/_node-agent.tpl`: `no_proxy_envar_list` is now threaded through to `node-agent.sbomScannerContainer` the same way it already is for `node-agent.container`, and the call site update (`nodeAgent.podSpec` → `node-agent.sbomScannerContainer`) passes it correctly.
- `kubevuln/deployment.yaml`: `$no_proxy_envar_list` is defined at file scope (top of the template) so it's in scope for the new sbom-scanner block; the added `HTTPS_PROXY`/`no_proxy` env entries are an exact mirror of the existing block for the main `kubevuln` container.
- New snapshot tests assert the right container index for both the node-agent daemonset and the kubevuln deployment, and match the existing test style in this file.
- `helm unittest` coverage (60 tests / 996 snapshots + 2 new) is consistent with the PR description.

Holding off on approval for one reason: the `allowlist-drift` check is failing — the new `sbom-scanner` env vars (`HTTPS_PROXY`, `no_proxy`) aren't in the approved GKE Autopilot allowlist (`armo-kubescape-node-agent-1.40-v2`), so on GKE Autopilot this node-agent spec would be rejected by GKE Warden until Google approves an updated allowlist. I see this is already being tracked (cc @slashben re: new autopilot submission) — once the Gerrit allowlist update lands and a maintainer adds `allowlist-drift-ack`, this is good to merge from my side.

Issue #531, opened 2024-11-03
Kubescape in ArgoCD always get out of sync (`divisor: '0'`)


mogaal:
# Description

When running kubescape helm chart in ArgoCD 
# Environment

OS: K8S (EKS)
Version: Helm Chart 1.22.0
 
# Steps To Reproduce

Install using ArgoCD as below:

```
---
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: kubescape
  namespace: argocd
spec:
  project: base-infra
  destination:
    namespace: kubescape
    server: https://kubernetes.default.svc
  sources:
    - repoURL: https://kubescape.github.io/helm-charts/
      chart: "kubescape-operator"
      targetRevision: 1.22.0
      helm:
        releaseName: kubescape
  syncPolicy:
   automated:
     prune: true
     selfHeal: true
   syncOptions:
    - CreateNamespace=true
    - ServerSideApply=true
```

# Expected behavior

All resources should keep the sync 

# Actual Behavior

It gets out of sync constantly, as you can see in the image below:

<img width="1676" alt="Screenshot 2024-11-03 at 17 09 30" src="https://github.com/user-attachments/assets/b304d9d4-2d7d-459c-b530-0afcb0f59e68">

The way how I "fixed" it is by telling ArgoCD to ignore those fields:

```
  ignoreDifferences:
  - group: apps
    kind: Deployment
    jqPathExpressions:
      - .spec.template.spec.containers[].env[].valueFrom.resourceFieldRef.divisor
  - group: apps
    kind: DaemonSet
    jqPathExpressions:
      - .spec.template.spec.containers[].env[].valueFrom.resourceFieldRef.divisor
```
 
# Additional context

N/A

matthyx:
@mogaal can you try with this https://github.com/kubescape/helm-charts/blob/main/ArgoCDApplication.yaml

PR #900, opened 2026-08-13
fix(storage): omit the empty annotations key on the seccompprofiles CRD

ahauserv:
## Overview

The seccompprofiles CRD template always writes an `annotations:` key. The helper that fills it only produces anything when `additionalAnnotations` is set, so by default the key stays empty:
https://github.com/kubescape/helm-charts/blob/main/charts/kubescape-operator/templates/storage/seccompprofile-crd.yaml#L6-L7

YAML reads the bare key as `annotations: null`, and the API server strips a null metadata map on write. The rendered manifest never matches the stored object, so anything that compares the two (`kubectl diff`, GitOps reconcilers, drift detection) reports this one field of this one object as different, indefinitely:

```
  metadata:
-   annotations: null
    name: seccompprofiles.kubescape.io
```

This wraps the block in `with`, so the key is rendered only when there is something to put in it. Annotations set via `additionalAnnotations` render exactly as before.

## How to Test

New tests in `tests/seccompprofile_crd_test.yaml` assert both directions — key omitted when `additionalAnnotations` is unset, annotations carried when it is set. The 11 snapshot changes are exactly the removed null key.

## Related issues/PRs:

* Related: #531 was the same class from a different field — a rendered value (`divisor: '0'`) the API server normalizes, so comparators never converge.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of annotations in the SeccompProfile resource, ensuring configured annotations are rendered correctly and omitted when not set.

* **Tests**
  * Added coverage for annotation rendering and omission scenarios.

### DIFF
--- a/charts/kubescape-operator/templates/storage/seccompprofile-crd.yaml
+++ b/charts/kubescape-operator/templates/storage/seccompprofile-crd.yaml
@@ -3,8 +3,10 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: seccompprofiles.kubescape.io
+  {{- with include "kubescape-operator.annotations" (dict "Values" .Values) }}
   annotations:
-    {{- include "kubescape-operator.annotations" (dict "Values" .Values) | nindent 4 }}
+    {{- . | nindent 4 }}
+  {{- end }}
   labels:
     {{- include "kubescape-operator.labels" (dict "Chart" .Chart "Release" .Release "Values" .Values "app" "seccompprofile" "tier" .Values.global.namespaceTier) | nindent 4 }}
 spec:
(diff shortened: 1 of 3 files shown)


matthyx (approved):
Reviewed the template change, the diff, and the updated helper output.

- The `with`-wrap correctly relies on Go templates treating an empty string from `kubescape-operator.annotations` as falsy, so the `annotations:` key is fully omitted when `additionalAnnotations` is unset — matching the drift-detection problem described in the PR body.
- When `additionalAnnotations` is set, the helper output and `nindent 4` call are unchanged from before, so that path renders identically to today.
- The 11 snapshot diffs are exactly the removed `annotations: null` lines, and the new `tests/seccompprofile_crd_test.yaml` covers both the omitted and populated cases.
- All CI checks (GitGuardian, DCO, allowlist-drift, pr-created) are green.

No blockers. One non-blocking note for a possible follow-up: the same unconditional `{{- include "kubescape-operator.annotations" ... | nindent 4 }}` pattern (which renders `annotations: null` by default) exists in ~25 other templates across this chart (e.g. `storage/pvc.yaml`, `operator/role.yaml`, `storage/service.yaml`, `node-agent/daemonset.yaml`, etc.). Not something this PR needs to fix given its scope (consistent with how #531 handled a different field), just flagging it as a candidate for a follow-up PR if drift-detection noise on those objects matters too.

Approving — good to merge.

PR #906, opened 2026-08-24
feat: add options to disable exec and open tracing

ystkfujii:
## Overview

This PR adds `execTracer` and `openTracer` configuration options for the node-agent.

Both options are enabled by default to preserve the current behavior. 
They can be disabled to prevent exec and open events from being included in SBOMSyft resources.

The README and Helm unit test snapshots have also been updated.

## Note

There seem to be other options that could be added, but it was unclear which ones should be included. 
Therefore, this PR only adds the options I need.

If you have any suggestions, I can add the other options in a separate PR.

https://github.com/kubescape/node-agent/blob/v0.3.197/docs/CONFIGURATION.md?plain=1#L399-L418

## Summary by CodeRabbit

* **New Features**
  * Added configuration options to enable or force-disable individual node-agent tracers, including exec and file-open event tracing.
  * Node-agent configuration now records disabled tracer settings.
  * All supported tracers remain enabled by default.
* **Documentation**
  * Added Helm upgrade guidance covering CRD schema changes and manual application steps.
  * Documented new risk-acceptance schema fields and per-tracer configuration options.

### DIFF
--- a/charts/kubescape-operator/README.md
+++ b/charts/kubescape-operator/README.md
@@ -191,6 +191,25 @@ However, we recommend that you give Kubescape no less than 500m CPU no matter th
 | operator.volumeMounts | object | `[]` | Additional volumeMounts for the web socket |
 | storage.hostNetwork | bool | `false` | Bind the storage APIServer to the host network. Required when using a custom CNI where the control plane cannot reach pod IPs |
 | nodeAgent.autoscaler.bottlerocketAutoDetect | bool | `true` | When the autoscaler is enabled, auto-detect AWS Bottlerocket nodes and set `seLinuxType: super_t` on the node-agent DaemonSet rendered for that node group, so you don't need to `--set nodeAgent.seLinuxType=super_t` manually |
+| nodeAgent.config.tracers.capSys | string | `enable` | Allow capSys tracing when required by enabled node-agent features. Set to disable to force-disable the capSys tracer |
+| nodeAgent.config.tracers.dns | string | `enable` | Allow dns tracing when required by enabled node-agent features. Set to disable to force-disable the dns tracer |
+| nodeAgent.config.tracers.exec | string | `enable` | Allow exec tracing when required by enabled node-agent features. Set to disable to force-disable the exec tracer |
+| nodeAgent.config.tracers.exit | string | `enable` | Allow exit tracing when required by enabled node-agent features. Set to disable to force-disable the exit tracer |
+| nodeAgent.config.tracers.fork | string | `enable` | Allow fork tracing when required by enabled node-agent features. Set to disable to force-disable the fork tracer |
+| nodeAgent.config.tracers.hardlink | string | `enable` | Allow hardlink tracing when required by enabled node-agent features. Set to disable to force-disable the hardlink tracer |
+| nodeAgent.config.tracers.http | string | `enable` | Allow HTTP tracing when required by enabled node-agent features. Set to disable to force-disable the HTTP tracer |
+| nodeAgent.config.tracers.iouring | string | `enable` | Allow io_uring tracing when required by enabled node-agent features. Set to disable to force-disable the io_uring tracer |
+| nodeAgent.config.tracers.network | string | `enable` | Allow network tracing when required by enabled node-agent features. Set to disable to force-disable the network tracer |
+| nodeAgent.config.tracers.open | string | `enable` | Allow file-open tracing when required by enabled node-agent features. Set to disable to force-disable the file-open tracer |
+| nodeAgent.config.tracers.ptrace | string | `enable` | Allow ptrace tracing when required by enabled node-agent features. Set to disable to force-disable the ptrace tracer |
+| nodeAgent.config.tracers.randomx | string | `enable` | Allow RandomX tracing when required by enabled node-agent features. Set to disable to force-disable the RandomX tracer |
+| nodeAgent.config.tracers.seccomp | string | `enable` | Allow seccomp tracing when required by enabled node-agent features. Set to disable to force-disable the seccomp tracer |
+| nodeAgent.config.tracers.ssh | string | `enable` | Allow SSH tracing when required by enabled node-agent features. Set to disable to force-disable the SSH tracer |
+| nodeAgent.config.tracers.symlink | string | `enable` | Allow symlink tracing when required by enabled node-agent features. Set to disable to force-disable the symlink tracer |
+| nodeAgent.config.tracers.kmod | string | `enable` | Allow kernel module tracing when required by enabled node-agent features. Set to disable to force-disable the kernel module tracer |
+| nodeAgent.config.tracers.unshare | string | `enable` | Allow unshare tracing when required by enabled node-agent features. Set to disable to force-disable the unshare tracer |
+| nodeAgent.config.tracers.bpf | string | `enable` | Allow BPF tracing when required by enabled node-agent features. Set to disable to force-disable the BPF tracer |
+| nodeAgent.config.tracers.top | string | `enable` | Allow top tracing when required by enabled node-agent features. Set to disable to force-disable the top tracer |
 | prometheusExporter.serviceMonitor.enabled | bool | `false` | enable/disable service monitor for prometheus-exporter integration |
 | awsIamRoleArn | string | `nil` | AWS IAM arn role |
 | cloudProviderMetadata.secretRef.name | string | `nil` | secret name to define values for the provider's metadata |
--- a/charts/kubescape-operator/templates/node-agent/configmap.yaml
+++ b/charts/kubescape-operator/templates/node-agent/configmap.yaml
@@ -69,6 +69,25 @@ data:
         },
         "alertDeduplication": {
             "bypass": {{ .Values.nodeAgent.config.alertDeduplication.bypass }}
-        }
+        },
+        "dCapSys": {{ eq .Values.nodeAgent.config.tracers.capSys "disable" }},
+        "dDns": {{ eq .Values.nodeAgent.config.tracers.dns "disable" }},
+        "dExec": {{ eq .Values.nodeAgent.config.tracers.exec "disable" }},
+        "dExit": {{ eq .Values.nodeAgent.config.tracers.exit "disable" }},
+        "dFork": {{ eq .Values.nodeAgent.config.tracers.fork "disable" }},
+        "dHardlink": {{ eq .Values.nodeAgent.config.tracers.hardlink "disable" }},
+        "dHttp": {{ eq .Values.nodeAgent.config.tracers.http "disable" }},
+        "dIouring": {{ eq .Values.nodeAgent.config.tracers.iouring "disable" }},
+        "dNetwork": {{ eq .Values.nodeAgent.config.tracers.network "disable" }},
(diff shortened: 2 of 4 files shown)


ystkfujii:
@matthyx 
Could you review this when you have time?

matthyx (dismissed):
Verified against node-agent's `dExec`/`dOpen` debug flags (both default `false` = tracer enabled, per [CONFIGURATION.md](https://github.com/kubescape/node-agent/blob/v0.3.197/docs/CONFIGURATION.md#debug-flags)): with the new `execTracer`/`openTracer` values defaulting to `enable`, `{{ eq ... "disable" }}` correctly renders `false` for both flags, preserving current behavior. Setting either to `disable` correctly turns the corresponding tracer off.

The `eq .Values.X "enable"`/`"disable"` pattern matches existing toggles in this same template (`prometheusExporter`, `hostMalwareSensor`, `hostNetworkSensor`, `runtimeDetection`, `relevancy`), README and snapshot tests are updated consistently, and CI is green. No blockers.

CodeRabbit's nitpick (add explicit Helm unit test cases for the `disable` value) is a reasonable optional follow-up, not a merge blocker.

matthyx:
@ystkfujii I'm curious to know what's your use case?

ystkfujii:
@matthyx 

Thank you for the quick review! 🚀

For our use case, we need to associate SBOMs with workloads. 
To do this, we generate SBOMSyft and ContainerProfile resources.

However, the `execs` and `opens` fields may contain sensitive information. 
I think it would be difficult to completely prevent such information from being recorded.

For example, these fields are populated when deploying Argo CD Redis using its default Helm chart.

https://github.com/argoproj/argo-helm/blob/50bc2fc864fe63eebc523c46f8baf33bf1e64882/charts/argo-cd/templates/redis/health-configmap.yaml#L14

We do not need this data, so we want to prevent any information from being recorded in these fields.

matthyx:
what about setting `  runtimeObservability: disable` ? if you're not interested in runtime info?

ystkfujii:
We are not interested in runtime information, but `runtimeObservability` is currently enabled in our configuration.
However, I am not sure whether `SBOMSyft` and `ContainerProfile` resources will still be generated if `runtimeObservability` is disabled.

My understanding is:

- `applicationProfileServiceEnabled` must be `true` to generate `ContainerProfile` resources.
  - FYI: https://github.com/kubescape/node-agent/blob/6fb44444703fdcb9fe64f665df52211d93067ffc/cmd/main.go#L285
- Either `runtimeObservability` or `relevancy` must be enabled for `applicationProfileServiceEnabled` to be `true`.
  - FYI: https://github.com/kubescape/helm-charts/blob/39a98fe8c24fd3769c03b00429d4a33842505e81/charts/kubescape-operator/templates/node-agent/configmap.yaml#L24

Is my understanding correct?
If so, would it be better to disable `runtimeObservability` and enable `relevancy`?

matthyx:
But why do you need ContainerProfiles if you're not interested in opens and execs? Also relevancy won't work if you don't have execs and opens...

ystkfujii:
> why do you need ContainerProfiles

I want to associate SBOMs with workloads. 
I only need the association, so I don't specifically need ContainerProfile resources.
ContainerProfile contains the image ID associated with the SBOMSyft name.

matthyx:
ok got you
cc @slashben what do you think?

slashben:
@ystkfujii I would like to suggest a different approach here.

1. If we decide to put levers controlling which eBPF program/gadgets are enabled, I don't want to make it open/exec specific. We would need something that controls all gadgets (so someone else can decide to disable the DNS gadget, for example)
2. I think it is a waste of effort and time to use ContainerProfiles for the association to ImageID, I'd much invest the time to include the ImageID in the SBOMSyft directly (should be a simple change on node-agent/kubevuln side, @matthyx to confirm)

I really want to solve your problem fast, but I want it to be aligned with the project

ystkfujii:
@slashben 
Thank you for considering my use case and suggesting an alternative approach 🙇 
My goal is to associate SBOMs with workloads:

`workload ← image ID → SBOMSyft`

For example, I want to identify which Pod are associated with each SBOMSyft.

Adding the image ID to SBOMSyft solves the `image ID → SBOMSyft` part, but I think I would still need a way to map workload to image IDs. I currently use ContainerProfile for that mapping.

> I really want to solve your problem fast, but I want it to be aligned with the project

I agree that the solution should be aligned with the project. 
If there is a better way to provide this mapping without runtime tracing, I would be happy to use it and help implement the preferred approach 🤝

slashben:
@ystkfujii I think you can build `workload ← image ID → SBOMSyft` today with no chart change and no runtime tracing. Two objects already carry both halves.

1. SBOMSyft already has the image ID

It's there as an annotation (exact) and as a label (sanitized, digest truncated to 32 hex chars for the 63-char limit):

```bash
kubectl get sbomsyfts -n kubescape <name> \
  -o jsonpath='{.metadata.annotations.kubescape\.io/image-id}'
# docker.io/library/nginx@sha256:6784fb0834aa7dbbe12e3d7471e69c290df3e6ba810dc38b34ae33d3c1c05f7d
```

Note: match on the annotation, the label is lossy.

2. `workload → image ID` comes straight from the Pod status

This is Kubernetes' own resolved digest, byte-identical to the annotation above, so ContainerProfile isn't needed for the mapping:

```bash
kubectl get pods -A -o json \
| jq -r '.items[] as $p
         | ($p.metadata.ownerReferences[0] // {kind:"Pod",name:$p.metadata.name}) as $o
         | $p.status.containerStatuses[]?
         | [.imageID, $p.metadata.namespace+"/"+$o.kind+"/"+$o.name, .name] | @tsv' \
| sort > wl.tsv

kubectl get sbomsyfts -A -o json \
| jq -r '.items[] | [.metadata.annotations["kubescape.io/image-id"],
                     .metadata.namespace+"/"+.metadata.name] | @tsv' \
| sort > sbom.tsv

join -t $'\t' sbom.tsv wl.tsv
```

On a fresh kind cluster with `kubectl create deployment nginx-test --image=nginx:1.27`, that gives me:

```
docker.io/library/nginx@sha256:6784fb…	kubescape/docker.io-library-nginx-1.27-c05f7d	default/ReplicaSet/nginx-test-f7b95f668	nginx
```

`kubectl get sbomsyfts` LIST returns metadata only (~5 KB for two SBOMs, vs ~1 MB for one full object), so joining client-side stays cheap at scale.

**If you'd rather stay inside Kubescape objects**

`VulnerabilityManifestSummary` carries workload identity **and** the image ID on the same object, and it comes from kubevuln (static scan) rather than node-agent tracing:

```bash
kubectl get vulnerabilitymanifestsummaries -A -o json \
| jq -r '.items[].metadata | [.annotations["kubescape.io/wlid"],
                              .labels["kubescape.io/workload-container-name"],
                              .annotations["kubescape.io/image-id"]] | @tsv'
# wlid://cluster-kind-kind/namespace-default/deployment-nginx-test	nginx	docker.io/library/nginx@sha256:6784fb…
```

One caveat before you build on the labels

The storage aggregated API server currently ignores label selectors on these types, a deliberately bogus selector returns *everything*, not zero:

```bash
kubectl get sbomsyfts -n kubescape -l kubescape.io/image-id=doesnotexist -o name
# returns ALL SBOMSyfts
```

So the join has to be client-side for now. That may deserve a separate issue, and it's arguably what would make the `kubescape.io/image-id` label genuinely useful.

Does this cover your use case, or is there something the client-side join doesn't give you — server-side filtering, or the mapping surfaced directly on the SBOMSyft?

ystkfujii:
Thanks for the detailed explanation.

I had also considered using the Pod status. In our current setup, using the resources we already generate would be a smaller change for us than introducing a client-side join.

I had not looked into `VulnerabilityManifestSummary`, so thank you for pointing that out.

After checking it briefly, my understanding is that using it would require enabling additional components such as the operator and kubevuln in our setup. That may be more operational overhead than we need just for this association.

That said, If using `VulnerabilityManifestSummary` is the preferred Kubescape-native approach for this use case,  I think it is worth considering. This discussion made me realize that I may need to reconsider how to solve my original use case. If further discussion is needed, I’ll open a separate issue 🙏

---

Separately from my specific use case, I still think there may be value in making tracer configuration available through the Helm chart.

> If we decide to put levers controlling which eBPF program/gadgets are enabled, I don't want to make it open/exec specific. We would need something that controls all gadgets (so someone else can decide to disable the DNS gadget, for example)

I agree that this probably should not be specific to `exec` and `open`. Would it make sense to provide a generic way to enable or disable the supported tracers/gadgets through the Helm chart, while keeping the current behavior as the default?

For example:

```yaml
nodeAgent:
  config:
    disabledTracers:
      - exec
      - open
```

or 

```yaml
nodeAgent:
  config:
    tracers:
      exec: disable
      open: disable
```

If this direction fits the project better, I would be happy to adjust the implementation accordingly.
thanks!

matthyx:
we already have "hidden" configs to disable individual tracers: https://github.com/kubescape/node-agent/blob/main/pkg/config/config.go#L64-L83

ystkfujii:
@matthyx 

Yes, that matches my understanding.

I was mainly thinking about how we should expose the existing `d*` options through Helm values.

For example:

```yaml
nodeAgent:
  config:
    disabledTracers:
      - exec
      - open
      - dns
      - ...
```

or:

```yaml
nodeAgent:
  config:
    tracers:
      exec: disable
      open: disable
      dns: disable
      ...
```

or even: 

```yaml
nodeAgent:
  config:
    dExec: true
    dOpen: true
    dDns: true
    ...
```

My preference is the second option, for example `nodeAgent.config.tracers.exec: disable`.

matthyx:
@ystkfujii sorry for the delay, let's implement your second option... do you have time to work on it or should I take over? (I want to release a chart soon-ish)

matthyx (approved):
Re-reviewed after the 2026-09-02 push (my prior approval was dismissed by the force-push that expanded scope from 2 tracers to all 19).

Verified the new commit set against node-agent's [CONFIGURATION.md Debug Flags](https://github.com/kubescape/node-agent/blob/v0.3.197/docs/CONFIGURATION.md#debug-flags) (v0.3.197): all 19 `dX` fields rendered in `configmap.yaml` (`dCapSys`, `dDns`, `dExec`, `dExit`, `dFork`, `dHardlink`, `dHttp`, `dIouring`, `dNetwork`, `dOpen`, `dPtrace`, `dRandomx`, `dSeccomp`, `dSsh`, `dSymlink`, `dKmod`, `dUnshare`, `dBpf`, `dTop`) map 1:1 to `values.yaml`'s `nodeAgent.config.tracers.*` keys, each defaulting to `enable` (preserving current behavior). The `eq .Values.X "disable"` pattern matches existing toggles in this template (`prometheusExporter`, `hostMalwareSensor`, etc.), README table and Helm snapshot tests are updated consistently, and CI is green.

No blockers. CodeRabbit's two "outside diff" README comments (CRD `--force-conflicts` / version pinning) are pre-existing content unrelated to this diff, not introduced by this PR. Its inline nitpick (strict-enum validation on `disable`/`enable` so typos fail render instead of silently enabling) is a reasonable optional follow-up, consistent with how the rest of the file already handles this pattern — not a merge blocker.

matthyx:
@ystkfujii thanks for the quick response

PR #911, opened 2026-08-26
feat(node-agent): add a size cap for the host malware sensor

AlonLiwsky:
## Summary

Adds `nodeAgent.config.hostMalwareSensorMaxFileSizeBytes`, which reaches the agent as
`hostMalwareSensorMaxFileSizeBytes` in `config.json`. Default `0`, which means no limit and is
exactly today's behaviour.

The host malware sensor hashes executed and opened files. An A/B benchmark measured its CPU cost
as **proportional to the bytes it hashes, with no fixed part** — about +32 mCPU per MB/s — and
two of its paths reach the hashing function with no size test of their own. So an operator
running the sensor on a busy node has no way to bound what it costs. This key is that bound.

## Ticket

None — this repository has no ticket link. Tracked internally as part of the file-hash detection
GA work.

## Changes

- `values.yaml`: new `nodeAgent.config.hostMalwareSensorMaxFileSizeBytes`, default `0`.
- `templates/node-agent/configmap.yaml`: render it into `config.json`.
- Snapshots regenerated.

### Why the two uncapped paths exist

1. An **executed** file is never size-checked. Both file filters return true immediately for an
   execution, before any size test.
2. An **opened** file whose extension is on the sensor's software-component list short-circuits
   ahead of the 250 MB filter. `.jar` is on that list, so a JVM reading large archives at
   start-up is hashed in full.

It is off by default because capping trades detection coverage for cost: a file over the cap is
not hashed, so a very large malicious binary is not detected.

## Testing

`helm template` renders the key into `config.json` with the default and with an override.
`helm unittest` in the same container image CI uses: 87 tests pass, 996 snapshots pass. The only
snapshot content change is the new key and the node-agent config checksums.

AI-skills: armosec-shared-rules:agent-dispatch-policy


AlonLiwsky:
Folded into #910. The two changes touch the same values file and the same snapshot, so keeping them apart only bought a rebase. #910 now carries the sidecar removal, the capability change and this size cap as three commits.

PR #912, opened 2026-08-28
fix: pin runAsGroup wherever the chart pins runAsUser

ahauserv:
The chart fails its own **C-0013 (non-root containers)** on almost every component it deploys.

The rule resolves `runAsGroup` from the container context, then the pod context, then defaults to 0. No `securityContext` in this chart sets it except the autoupdater CronJob, even where `runAsUser` is set right next to it.

PR #897 closed the same self-scan gap for capabilities.

This adds the matching `runAsGroup` to all 13 sites, each taking the `runAsUser` already
on that context:

| Site | Context | Value |
|---|---|---|
| `kubescape`, `kubevuln`, `operator`, `prometheus-exporter`, `storage`, `synchronizer` | pod | 65532 |
| `operator` ×2, `storage` ×2 | certgen init container | 65532 |
| `grype-offline-db`, `kubescape-scheduler`, `kubevuln-scheduler` | CronJob container | 100 |

The node-agent keeps `runAsUser: 0` and is untouched. `fsGroup` does not cover this:
it adds a supplementary group and leaves the primary GID alone.

Where the pod sets `runAsUser: 65532` the group already resolves to 65532 from the image, so pinning is a formality. Where a *container* sets `runAsUser: 100`, the uid overrides the image's user, nothing resolves the group, and it falls back to root. `/proc/<pid>/status` for a running `kubescape-scheduler`:

```
Name:   http_request
Uid:    100     100     100     100
Gid:    0       0       0       0
Groups: 0
```

`kubevuln-scheduler` shares that image and context; `grype-offline-db`'s rollout has the
same `runAsUser: 100` over a different image. So three CronJob containers run as group
root today.

`runAsUser: 100` matches no user in the `http-request` image, whose `USER` is `nonroot`
(65532). I kept each context internally consistent with `runAsGroup: 100`, but if the
intent was the image's own user, it should be 65532 in both fields. Happy to switch.

The `grype-offline-db` Deployment sets no `runAsUser` at all, so there is no user there to
match a group to, and it is left alone.

## How to Test

```
helm unittest charts/kubescape-operator/
```

105 tests and 993 snapshots pass. The snapshot diff is 105 added `runAsGroup` lines, no removals and nothing else.

## Related issues/PRs:

* Follows #897, which closed the capabilities half of the same self-scan gap.
* Related: #740 asked for configurable securityContexts.
  As with #897, the hardcoded contexts were nearly complete already, so finishing them closes these rows without a knob.

- [ ] If it is a core feature, I have added thorough tests. — not a core feature; the
  existing snapshot suite covers every added line

## Summary by CodeRabbit

- **Security Enhancements**
  - Added explicit group IDs to operator components, scheduled jobs, and supporting services.
  - Improved consistency between configured user and group identities.
  - Ensured workloads and certificate-generation processes run with defined non-root group permissions.

### DIFF
--- a/charts/kubescape-operator/templates/grype-offline-db/cronjob.yaml
+++ b/charts/kubescape-operator/templates/grype-offline-db/cronjob.yaml
@@ -54,6 +54,7 @@ spec:
               readOnlyRootFilesystem: true
               runAsNonRoot: true
               runAsUser: 100
+              runAsGroup: 100
               capabilities:
                 drop:
                   - ALL
--- a/charts/kubescape-operator/templates/kubescape-scheduler/cronjob.yaml
+++ b/charts/kubescape-operator/templates/kubescape-scheduler/cronjob.yaml
@@ -56,6 +56,7 @@ spec:
               readOnlyRootFilesystem: true
               runAsNonRoot: true
               runAsUser: 100
+              runAsGroup: 100
               capabilities:
                 drop:
                   - ALL
--- a/charts/kubescape-operator/templates/kubescape/deployment.yaml
+++ b/charts/kubescape-operator/templates/kubescape/deployment.yaml
@@ -58,6 +58,7 @@ spec:
         seccompProfile:
           type: RuntimeDefault
         runAsUser: 65532
+        runAsGroup: 65532
         fsGroup: 65532
       containers:
       - name: kubescape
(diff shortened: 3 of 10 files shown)


matthyx (approved):
LGTM. Every `runAsGroup` addition sits next to an existing `runAsUser` and takes its value, so this is purely closing the chart's self-scan gap (mirrors #897 for capabilities) with no behavior change for already-conformant contexts. Snapshot diff is exactly the 105 added lines described in the PR body, no removals.

One non-blocking note on the point you already flagged: for the three CronJob containers pinned at `runAsUser: 100` (grype-offline-db, kubescape-scheduler, kubevuln-scheduler), matching `runAsGroup: 100` keeps things internally consistent but doesn't fix the underlying mismatch with the image's actual `nonroot` (65532) user — that's pre-existing and out of scope here, so fine to follow up separately if it's worth changing.

PR #914, opened 2026-09-01
fix: bump kubescape scanner to v4.0.12 to unbreak standalone configuration scanning

slashben:
Fixes configuration scanning on standalone (no ARMO account) installs, which is currently **completely broken**: no `WorkloadConfigurationScan` or `ConfigurationScanSummary` object is ever persisted.

Reported in kubescape/kubescape#3707.

## The bug

The chart pins `quay.io/kubescape/kubescape:v4.0.11`. On a standalone install `accountID` is empty, so `Submit()` always takes the account-ID generation path, and in v4.0.11 `updateConfigFile()` hard-fails on chmod:

```go
if err := os.Chmod(dir, 0700); err != nil {
    return err
}
```

The scanner runs `runAsNonRoot: true` + `readOnlyRootFilesystem: true`, and `/home/nonroot/.kubescape` is a **root-owned `emptyDir` subPath mount** — a non-root process cannot chmod it. The `EPERM` propagates out and aborts the scan *after* it has already succeeded:

```
info  Scan results saved              filename=results/03dca245-....json
error failed to generate account ID   reason=chmod /home/nonroot/.kubescape: operation not permitted
error scanning failed                 error=failed to scan. reason: 'chmod ...'
```

Deterministic — every scan trigger observed failed identically. Vulnerability scanning and runtime observability are unaffected, which is what makes it easy to miss:

| resource | count |
|---|---|
| `vulnerabilitymanifests` | 6 |
| `applicationprofiles` | 4 |
| `networkneighborhoods` | 4 |
| `workloadconfigurationscans` | **0** |
| `configurationscansummaries` | **0** |

## Why the tag bump is the whole fix

`v4.0.12` already contains the fix — chmod is best-effort (logged as a warning) and the config is written via temp-file + rename, so no chmod on the target is needed at all. Only the pinned tag keeps it from users. The running v4.0.11 scanner even logs:

```
warn  current version '4.0.11' is not updated to the latest release: 'v4.0.12'
```

Because the account-ID path only runs when `accountID` is empty, this affects **standalone installs only** — likely why cloud-connected testing does not surface it.

## Changes

- `values.yaml`: `kubescape.image.tag` `v4.0.11` → `v4.0.12` (no other image touched)
- `Chart.yaml`: patch bump `1.40.3` → `1.40.4` (version + appVersion)
- Snapshots regenerated with `helm unittest -u`

## Verification

`helm unittest charts/kubescape-operator` — **105 tests / 993 snapshots pass**.

Found while verifying the Kubescape tool provider in `kagent-dev/tools` (`pkg/kubescape`), where this also silently disables the `kubescape_list_configuration_scans` and `kubescape_get_configuration_scan` tools.

Environment: DOKS k8s 1.34.10, Debian 13, kernel 6.12.96, 3 × s-4vcpu-8gb, chart 1.40.3 defaults + `continuousScan=enable`.

AI-skills: superpowers:using-git-worktrees

## Summary by CodeRabbit

* **Updates**
  * Updated the Kubescape Operator chart and application versions to 1.40.4.
  * Updated the Kubescape scanner image to version 4.0.12.

### DIFF
--- a/charts/kubescape-operator/Chart.yaml
+++ b/charts/kubescape-operator/Chart.yaml
@@ -15,14 +15,14 @@ kubeVersion: ">=1.28.0-0"
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
 
-version: 1.40.3
+version: 1.40.4
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
 # It is recommended to use it with quotes.
 
-appVersion: 1.40.3
+appVersion: 1.40.4
 
 maintainers:
 - name: Ben Hirschberg
--- a/charts/kubescape-operator/tests/__snapshot__/snapshot_test.yaml.snap
+++ b/charts/kubescape-operator/tests/__snapshot__/snapshot_test.yaml.snap
@@ -1,7 +1,7 @@
 all capabilities:
   1: |+
     raw: |+
-      Thank you for installing kubescape-operator version 1.40.3.
+      Thank you for installing kubescape-operator version 1.40.4.
       View your cluster's configuration scanning schedule:
       > kubectl -n kubescape get cj kubescape-scheduler -o=jsonpath='{.metadata.name}{"\t"}{.spec.schedule}{"\n"}'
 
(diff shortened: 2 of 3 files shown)


matthyx (approved):
LGTM! Bumped kubescape scanner image to v4.0.12 and chart version to 1.40.4 with updated test snapshots.

PR #916, opened 2026-09-05
feat(storage,node-agent): expose profile size caps and add GOMEMLIMIT headroom

### LINKED ISSUE #885: Expose maxTsProfileSize and maxApplicationProfileSize; give storage GOMEMLIMIT headroom
## Summary

Two of the three caps that govern container profile size are not exposed in `values.yaml`, so a customer hitting a size rejection cannot be unblocked without a new image. Separately, storage pins `GOMEMLIMIT` to `limits.memory` exactly, which makes raising the size cap risky.

Surfaced while investigating a customer (ISO Gruppe) on Rancher/RKE2 hitting `request entity too large` (`contentLength=3231797`, `maxObjectSize=2500000`) for ContainerProfiles on `cattle-system` workloads. See kubescape/node-agent#865 and kubescape/storage#350.

## 1. Unexposed caps

| Cap | Default | Unit | Chart-exposed? |
|---|---|---|---|
| node-agent `maxTsProfileSize` | 2 MiB | Go heap bytes | ❌ |
| storage `kindQueues.containerprofiles.maxObjectSize` | 2,500,000 | wire bytes | ✅ `storage.kindQueues` |
| storage `maxApplicationProfileSize` (also used as `MaxContainerProfileSize`) | 40,000 | item count | ❌ |
| storage `maxNetworkNeighborhoodSize` | 40,000 | item count | ❌ |

Only `maxObjectSize` can be tuned. Because the node-agent split target and the storage rejection threshold are in **different units** with only ~19% nominal headroom, raising just the storage cap is a partial workaround — the node-agent side needs to move too.

### Requested

- Add `nodeAgent.config.maxTsProfileSize` and template it into `templates/node-agent/configmap.yaml`.
- Add `storage.maxApplicationProfileSize` and `storage.maxNetworkNeighborhoodSize` and template them into `templates/storage/configmap.yaml`. Both are already viper keys in storage; the configmap simply does not emit them.
- Regenerate `tests/__snapshot__/snapshot_test.yaml.snap`.

## 2. storage GOMEMLIMIT has no headroom

[templates/storage/deployment.yaml:142-146](https://github.com/kubescape/helm-charts/blob/main/charts/kubescape-operator/templates/storage/deployment.yaml#L142-L146):

```yaml
- name: GOMEMLIMIT
  valueFrom:
    resourceFieldRef:
      resource: limits.memory
      divisor: '1'
```

`GOMEMLIMIT` equals the cgroup limit exactly, so the Go runtime only starts trying to stay under the limit at the point where the kernel is already entitled to OOM-kill. Raising `maxObjectSize` without also raising storage memory limits risks the OOM crashloop shape we already saw on one prod env.

A helper for this already exists and is used by kubevuln and node-agent:

```
{{ include "kubescape-operator.gomemlimit" (dict "memory" .Values.storage.resources.limits.memory "percentage" .Values.storage.gomemlimitPercentage) }}
```

### Requested

Add `storage.gomemlimitPercentage` (default `0.8`, matching `kubevuln` and `nodeAgent`) and switch storage to the helper. `kubescape`, `operator`, `prometheus-exporter` and `synchronizer` are in the same 100% group and are probably worth the same treatment, though storage is the one with a demonstrated OOM history.

## Related

- kubescape/node-agent#865 — node-agent retries the rejected chunk forever
- kubescape/storage#350 — cap alignment and units

farhan-beg:
## What

Exposes the two untunable ContainerProfile size caps through the chart and gives storage's Go runtime GC headroom:

- `nodeAgent.config.maxTsProfileSize` (bytes, default 2097152 = node-agent viper default) → rendered into node-agent `config.json`
- `storage.maxContainerProfileSize` (item count, default 40000 = storage viper default) → rendered into storage `config.json`
- `storage.gomemlimitPercentage` (default 0.8) → storage GOMEMLIMIT now computed via the existing `kubescape-operator.gomemlimit` helper instead of raw `limits.memory` (1:1 mapping gives the GC zero headroom, the OOM-crashloop shape from the linked customer incident)

## Why

A customer (ISO Gruppe, Rancher/RKE2) hit `request entity too large` on ContainerProfiles. The caps live in three different units (heap bytes / wire bytes / item count) with ~19% nominal headroom, and only `maxObjectSize` was chart-exposed, so operators could not unblock without a new container image. Raising the caps safely also requires GOMEMLIMIT headroom, hence part 2. Details: #885, kubescape/node-agent#865 (Go-side fix shipped in node-agent#866, chart side remained open), kubescape/storage#350.

## Note on issue key names

The issue asks for `maxApplicationProfileSize` / `maxNetworkNeighborhoodSize`, but current storage code (`pkg/config/config.go`) defines `maxContainerProfileSize` (viper default 40000) and has no neighborhood-size key, so this PR exposes the real keys. Happy to adjust if maintainers prefer upstream key additions first.

## Testing

- `helmunittest/helm-unittest` docker image (same as CI): 108/108 tests, 1060/1060 snapshots green
- New snapshot case `storage profile size caps and GOMEMLIMIT headroom` with non-default values (2000Mi limit x 0.7 → renders `GOMEMLIMIT: 1400MiB`, caps render as 1e+06 / 50000), proving end-to-end wiring rather than defaults

Closes #885

## Summary by CodeRabbit

- **New Features**
  - Added configurable limits for node-agent time-series profiles and storage container profiles.
  - Added a storage memory-headroom percentage setting for controlling memory usage.

- **Bug Fixes**
  - Storage now applies its memory limit setting only when a memory limit is configured, helping prevent invalid or unintended memory configuration.

- **Documentation**
  - Added default values and configuration entries for the new profile-size and memory-headroom settings.

### DIFF
--- a/charts/kubescape-operator/templates/node-agent/configmap.yaml
+++ b/charts/kubescape-operator/templates/node-agent/configmap.yaml
@@ -37,6 +37,7 @@ data:
         "networkStreamingEnabled": {{ $gates.networkStreamingEnabled }},
         "maxImageSize": {{ .Values.kubevuln.config.maxImageSize }},
         "maxSBOMSize": {{ .Values.kubevuln.config.maxSBOMSize }},
+        "maxTsProfileSize": {{ .Values.nodeAgent.config.maxTsProfileSize }},
         "sbomGenerationEnabled": {{ eq .Values.capabilities.nodeSbomGeneration "enable" }},
         "enableEmbeddedSBOMs": {{ eq .Values.capabilities.scanEmbeddedSBOMs "enable" }},
         "seccompServiceEnabled": {{ eq .Values.capabilities.seccompProfileService "enable" }},
--- a/charts/kubescape-operator/templates/storage/configmap.yaml
+++ b/charts/kubescape-operator/templates/storage/configmap.yaml
@@ -23,6 +23,7 @@ data:
       "defaultQueueLength": {{ .Values.storage.defaultQueueLength }},
       "defaultWorkerCount": {{ .Values.storage.defaultWorkerCount }},
       "defaultMaxObjectSize": {{ .Values.storage.defaultMaxObjectSize }},
+      "maxContainerProfileSize": {{ .Values.storage.maxContainerProfileSize }},
       "queueManagerEnabled": {{ .Values.storage.queueManagerEnabled }},
       "kindQueues": {{ .Values.storage.kindQueues | toJson }},
       {{- if hasKey .Values.storage "singleWriterEnabled" }}
--- a/charts/kubescape-operator/templates/storage/deployment.yaml
+++ b/charts/kubescape-operator/templates/storage/deployment.yaml
@@ -149,11 +149,10 @@ spec:
             port: {{ .Values.storage.serverPort }}
             scheme: HTTPS
         env:
+          {{- if .Values.storage.resources.limits.memory }}
           - name: GOMEMLIMIT
-            valueFrom:
-              resourceFieldRef:
-                resource: limits.memory
-                divisor: '1'
+            value: {{ include "kubescape-operator.gomemlimit" (dict "memory" .Values.storage.resources.limits.memory "percentage" .Values.storage.gomemlimitPercentage) | quote }}
+          {{- end }}
           - name: KS_LOGGER_LEVEL
             value: "{{ .Values.logger.level }}"
           - name: KS_LOGGER_NAME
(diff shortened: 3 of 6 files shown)


matthyx (approved):
Reviewed the chart wiring against the current node-agent and storage config keys. The defaults preserve upstream behavior, storage gets GC headroom through the existing helper, and the non-default values are covered end-to-end by the regenerated snapshot. Local verification: 108/108 Helm tests and 1060/1060 snapshots pass; helm lint reports no failures.

PR #918, opened 2026-09-07
docs: fix reversed GC tuning guidance in kubevuln, storage, nodeAgent comments

farhan-beg:
Small follow-up to #917 while fixing the GC comment there I noticed kubevuln, storage and nodeAgent still have the old reversed wording ("raise toward 1.0 if GC is not aggressive enough", it's the other way around). Storage came in with #916 so it's outside that PR's diff, hence a separate PR.

Comment only, nothing renders differently.

### DIFF
--- a/charts/kubescape-operator/values.yaml
+++ b/charts/kubescape-operator/values.yaml
@@ -451,8 +451,8 @@ kubevuln:
   name: kubevuln
 
   # GOMEMLIMIT as a fraction of resources.limits.memory. 0.8 = 80%.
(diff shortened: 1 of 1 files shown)


matthyx (approved):
Verified all three affected guidance blocks. Raising gomemlimitPercentage moves GOMEMLIMIT toward the container limit and makes GC less aggressive, while lowering it gives more headroom and makes GC more aggressive. This comment-only correction is accurate and consistent with the follow-up in #917. No blockers.

PR #921, opened 2026-09-13
feat(operator): expose operator.rulesUpdateConfig in values.yaml

farhan-beg:
The operator reads `rulesUpdateConfig` from its config.json, and `templates/operator/configmap.yaml` already renders that block, but values.yaml never declared the keys. So the rules updater could only be switched on with `--set`, and nothing about it showed up in `helm show values`.

values.yaml now declares what the template already expects:

    rulesUpdateConfig:
      enabled: false
      interval: 5m

Default stays off, matching the operator's own viper default (config/config.go sets rulesUpdateConfig.enabled false and interval 5m), so existing renders only gain the block that was omitted before.

Added `tests/operator_rules_update_test.yaml` for the default and the enabled case.

Verification: `docker run --rm -v "$(pwd)":/apps helmunittest/helm-unittest charts/kubescape-operator/` gives 14 suites, 110 tests, 1060 snapshots, all green.

Signed-off-by: Farhan <farhanbeg30@proton.me>

## Summary by CodeRabbit

- **New Features**
  - Added Helm configuration for periodically pushing admission rules from Rule resources to node agents.
  - The feature is disabled by default and uses a 5-minute update interval.
  - Administrators can enable it and customize the interval, such as setting updates to every 30 minutes.

- **Tests**
  - Added Helm validation covering the default disabled state and custom interval configuration.


matthyx (approved):
Reviewed `3fa360a3c77a5dcfb6771282dd1e4d13ebe37abe` against `main` at `eca8cc9e61299c00206a9033c06e6e94980c66b2`.

This is needed as a discoverability/default-declaration improvement: the template has supported `operator.rulesUpdateConfig` since #739, but the keys are absent from the current chart defaults and default render. The new disabled/5m values match operator v0.2.169, and the existing template supplies the chart namespace. Minor wording nuance: users could already supply this map through a custom values file as well as `--set`.

No blockers found. The expected upgrade effect is a one-time operator rollout because the config checksum changes even though the updater remains disabled.

Validation: isolated full Helm unit suite (14 suites, 110 tests, 1,060 snapshots), `helm lint`, `git diff --check`, and explicit base/head renders all passed. The base omits the map by default but renders it when explicitly configured; the head exposes the defaults and renders both false/5m and true/30m cases. All current GitHub checks pass, the PR is mergeable, and there are no unresolved review threads.

Related-history search found merged #739, which introduced the template support without declaring defaults; it complements rather than supersedes this PR. No prior rejection was found. Search was bounded to GitHub issue/PR metadata and repository history using `rulesUpdateConfig`, updater/admission-rule phrases, affected template paths, and interval terms.

PR #685, opened 2025-05-02
WIP:  Use secret to pass provider's metadata

nikolaof:
## Overview
Currently cloudProviderMetadata parameters are forced to passed as plaintext values and some of them are kinda confidential. I would suggest to allow the helm-chart template to accept values from secrets.

## Related issues/PRs:

Resolved #683

## Checklist before requesting a review

put an [x] in the box to get it checked 

- [ ] My code follows the style guidelines of this project
- [x] I have commented on my code, particularly in hard-to-understand areas
- [x] I have performed a self-review of my code
- [ ] If it is a core feature, I have added thorough tests.
- [ ] New and existing unit tests pass locally with my changes

**Please open the PR against the `dev` branch (Unless the PR contains only documentation changes)**

-->

Issue #683, opened 2025-04-28
Pass cloudProviderMetadata as secrets


nikolaof:
## Overview
Currently `cloudProviderMetadata` parameters are forced to passed as plaintext values and some of them are kinda confidential. I would suggest to allow the helm-chart template to accept values from secrets.

## Problem
Expose of confidential information

## Solution
An approach like the following one could allow the usage of k8s secrets and plaintext values.

```
        {{- /* Check configuration to use a Secret if provided */}}
        {{- if .Values.cloudProviderMetadata.secretRef }}

          {{- if .Values.cloudProviderMetadata.secretRef.subscriptionIdKey }}
        - name: AZURE_SUBSCRIPTION_ID
          valueFrom:
            secretKeyRef:
              name: "{{ .Values.cloudProviderMetadata.secretRef.name }}"
              key: "{{ .Values.cloudProviderMetadata.secretRef.subscriptionIdKey }}"
          {{- end }}

          {{- if .Values.cloudProviderMetadata.secretRef.clientIdKey }}
        - name: AZURE_CLIENT_ID
          valueFrom:
            secretKeyRef:
              name: "{{ .Values.cloudProviderMetadata.secretRef.name }}"
              key: "{{ .Values.cloudProviderMetadata.secretRef.clientIdKey }}"
          {{- end }}

          {{- if .Values.cloudProviderMetadata.secretRef.clientSecretKey }}
        - name: AZURE_CLIENT_SECRET
          valueFrom:
            secretKeyRef:
              name: "{{ .Values.cloudProviderMetadata.secretRef.name }}"
              key: "{{ .Values.cloudProviderMetadata.secretRef.clientSecretKey }}"
          {{- end }}

          {{- if .Values.cloudProviderMetadata.secretRef.tenantIdKey }}
        - name: AZURE_TENANT_ID
          valueFrom:
            secretKeyRef:
              name: "{{ .Values.cloudProviderMetadata.secretRef.name }}"
              key: "{{ .Values.cloudProviderMetadata.secretRef.tenantIdKey }}"
          {{- end }}

          {{- if .Values.cloudProviderMetadata.secretRef.resourceGroupKey }}
        - name: AZURE_RESOURCE_GROUP
          valueFrom:
            secretKeyRef:
              name: "{{ .Values.cloudProviderMetadata.secretRef.name }}"
              key: "{{ .Values.cloudProviderMetadata.secretRef.resourceGroupKey }}"
          {{- end }}

[...]

PR #922, opened 2026-09-13
values: declare cloudProviderMetadata AKS and secretRef keys

farhan-beg:
The kubescape deployment template reads `cloudProviderMetadata.aksSubscriptionID`, `aksResourceGroup`, `aksClientID`, `aksClientSecret`, `aksTenantID` and the whole `secretRef` block, and README.md documents all of them, but values.yaml only declared `cloudRegion`, `awsIamRoleArn`, `gkeServiceAccount` and `gkeProject`. On AKS there was nothing to fill in and no way to find the key names from `helm show values`.

This also turns the commented `secretRef` example into declared keys. One thing worth knowing while reading the template: it picks the secret branch as soon as anything under `secretRef` is set, so configuring only `name` with no `*Key` mapping renders no metadata env vars at all. I kept that behaviour and documented it instead of changing how the template resolves, but I can follow up with a per-variable fallback (secret key first, inline value when that key is missing) if you prefer that direction.

`tests/cloud_provider_metadata_test.yaml` covers the inline AKS/GKE values, the secret path and the name-only case. No change to rendered defaults, snapshots untouched.

Signed-off-by: Farhan <farhanbeg30@proton.me>

## Summary by CodeRabbit

* **New Features**
  * Added configuration options for inline AKS cloud provider metadata.
  * Added support for sourcing cloud provider metadata from a referenced Secret using configurable key mappings.
  * Documented behavior for Secret-based metadata configuration, including cases where no metadata environment variables are rendered.

* **Tests**
  * Added Helm chart tests covering inline metadata, Secret-based metadata, and incomplete Secret configurations.


matthyx (commented):
Verdict: **needs clarification before approval** (reviewed `07270c16a56a6f0cf51ea1ff854001655216b38c` against `main` at `eca8cc9e61299c00206a9033c06e6e94980c66b2`).

The change is needed: on the base commit, `helm show values` omits the AKS fields and does not declare `secretRef`, although the deployment and README already support them. This PR makes those inputs discoverable without changing default rendering.

**MEDIUM — `charts/kubescape-operator/values.yaml:253` documents the wrong Secret contract.** It says all `*Key` fields are required, but `templates/kubescape/deployment.yaml:131-184` emits every mapping independently, and this PR's own Secret test omits `gkeProjectKey`. Is the intended contract provider-specific partial mappings (the current behavior), or every mapping? Please align the comment with the implementation. If partial mappings are intended, a precise version would be: a nonempty `secretRef` suppresses inline metadata environment variables; set `name` and only the key mappings needed for the selected provider. Requiring unrelated mappings can lead users to add a missing Secret key reference, which prevents the container from starting.

Related history: [#317](https://github.com/kubescape/helm-charts/pull/317) introduced AKS variables; [#683](https://github.com/kubescape/helm-charts/issues/683) and merged [#685](https://github.com/kubescape/helm-charts/pull/685) added the Secret path. Those discussions approved this direction; no prior rejection applies. Repository searches across PRs/issues for `cloudProviderMetadata`, `aksSubscriptionID`, `subscriptionIdKey`, `gkeProjectKey`, `secretRef`, and alternate “cloud/provider metadata” terms found no other superseding change. Search is limited to GitHub's indexed issue/PR corpus and 100 results per query.

Validation: full Helm unittest suite passed in a network-disabled, read-only container (14 suites, 111 tests, 1060 snapshots); `helm lint` had 0 failures (only the existing missing-local-dependencies warning); `git diff --check` passed; all reported checks for the reviewed head are green. The three new tests also pass when overlaid on the base commit, so they lock pre-existing template behavior rather than regress the values-discoverability gap. No runtime regression or architecture blocker was found beyond the documentation contract above.

farhan-beg:
Fair, the comment overstated it. Rewritten to match what the template actually does: a non-empty secretRef suppresses the inline metadata env vars, so it is `name` plus only the key mappings the provider you are on needs, not every mapping.

No template change, no change to rendered defaults. Also rebased onto main now that #921 landed.

matthyx (approved):
Re-reviewed `a86b6085ed8fc9a3eb35f4e142d4dfa1edf95692` against `main` at `a0c4b51e1dc60a3f34382d4ca0de602c83cea3d1`: **approved**.

The revised `values.yaml` comment now accurately describes the existing `secretRef` precedence and provider-specific mappings, resolving my previous concern. The change remains needed to expose already-supported AKS and Secret inputs through `helm show values`, preserves default rendering, and introduces no new template behavior.

I reviewed the new CodeRabbit thread. Adding direct assertions for every Secret mapping would improve failure localization, but it is non-blocking: the existing snapshot test already covers all five AKS Secret references, and the unchanged GKE mapping is statically consistent with the documented key. The focused suite still covers inline values, Secret-backed rendering, and name-only suppression.

Validation on this head: isolated network-disabled/read-only Helm unittest run passed 15 suites, 113 tests, and 1060 snapshots; `helm lint` had 0 failures (only the existing missing-local-dependencies warning); `git diff --check` passed; all reported required checks are green. Fresh related-history searches found no duplicate or superseding change; the previously identified #317 and #683/#685 history supports this direction and contains no rejected approach that applies here.

PR #925, opened 2026-09-18
fix(rbac): grant the operator Secrets only with kubevuln, pin certgen's get

ahauserv:
## Overview

Two RBAC narrowings, no behaviour change for installs that use the grants:

**`operator` Role — `secrets` only when `kubevuln` is enabled.**
The operator's Secret access exists for registry scanning: it lists and reads the credentials kubevuln scans registries with (`mainhandler/imageregistryhandler.go`, `vulnscan.go`) and deletes them with a registry scan's CronJob (`registryhandler.go`). It has no other reader. On a posture-only install (`capabilities.vulnerabilityScan: disable`) the grant is dead weight in the scanner's own namespace and shows up in the operator's own RBAC controls (C-0186 "Minimize access to secrets" flags the `operator` RoleBinding). The rule is now gated on `$components.kubevuln.enabled`, the same way the DaemonSet rule below it is gated on the node-agent autoscaler. `configmaps` and `cronjobs` stay unconditional — the scheduled-scan path uses them regardless of kubevuln.

**`storage-certgen` Role — pin `get` to the one Secret it reads.**
The `certgen-create`/`certgen-patch` init containers read back exactly the Secret they mint (`storage.certgen.secretName`); `get` now carries that `resourceNames`. `create` stays unpinned because a name is not known at authorization time and RBAC cannot restrict `create` by it.

Both were found by running kubescape's own RBAC controls against a posture-only install with `riskAcceptance` on: the two RoleBindings were the only Secret readers in the scanner's namespace without a reason to be one.

## How to Test

```sh
helm template ks charts/kubescape-operator --set clusterName=t --set capabilities.vulnerabilityScan=disable -s templates/operator/role.yaml   # no secrets rule
helm template ks charts/kubescape-operator --set clusterName=t --set capabilities.vulnerabilityScan=enable  -s templates/operator/role.yaml   # secrets rule present, verbs unchanged
helm template ks charts/kubescape-operator --set clusterName=t --set certificates.strategy=initContainer -s templates/storage/certgen/role.yaml # get pinned to storage-tls-init
helm unittest charts/kubescape-operator
```

Snapshots updated with `helm unittest -u`; the diff is the split rule (additions only) and the `resourceNames` on the certgen `get`.

## Related issues/PRs

- #242 introduced the operator's Secret access for registry scanning.
- #843 introduced the `initContainer` certificate strategy and its Role.

## Summary by CodeRabbit

- **Bug Fixes**
  - Restricted secret access to only when vulnerability scanning is enabled.
  - Limited certificate-generation read access to the designated certificate secret while preserving required secret creation capability.


matthyx (approved):
Reviewed `4c91f53` against `main` (`0e11c46`). Approving — no blockers.

**Need / correctness verified against the operator source (`kubescape/operator` main):**
- Every namespaced Secret API call in the operator lives on the kubevuln path: `mainhandler/vulnscan.go` (`loadRegistrySecret`, `getImageScanConfig`), `mainhandler/imageregistryhandler.go` (`getRegistryScanSecrets`) and `watcher/registryhandler.go` (create/delete via `secretGVR`). `RegistryCommandsHandler` is only started inside `if mainHandler.config.Components().Kubevuln.Enabled` in `mainhandler/handlerequests.go`, and `scanImage` / `scanRegistriesV2AndUpdateStatus` / `scanContainerProfile` all bail early with "kubevuln is not enabled". `Components().Kubevuln.Enabled` is fed by `components-configmap.yaml`, which uses the same `components` helper the Role now gates on, so chart and binary agree.
- The operator's admission-webhook certgen init containers use their own `operator/admission-webhook/role.yaml` (`get`/`create` on secrets), so they are unaffected by the operator Role change.
- `storage-certgen`: `scripts/certgen-create.sh` and `certgen-patch.sh` only `kubectl get secret "$SECRET_NAME"` where `$SECRET_NAME` is `storage.certgen.secretName` (renders `storage-tls-init`), plus one `create`. Pinning `get` with `resourceNames` and leaving `create` unpinned is correct — RBAC cannot restrict `create` by name.

**Validation run locally:**
- `helm template` for `vulnerabilityScan=disable` / `enable` and `certificates.strategy=initContainer` — output matches the PR description.
- `helm unittest charts/kubescape-operator` on the PR head: 15 suites / 111 tests / 1060 snapshots pass. Restoring the two `main` templates against the PR snapshot fails, so the snapshot does exercise the change.
- Ad-hoc `helm unittest` asserts (secrets rule absent with `vulnerabilityScan=disable`, present with `enable`, certgen `get` pinned to `storage-tls-init` + unpinned `create`) — all pass.
- Searched open/closed PRs and issues for overlapping work (secrets, rbac, certgen, C-0186, resourceNames); nothing duplicate or previously rejected. #242 / #843 are the origins as stated.

**Nonblocking suggestions:**
- No snapshot scenario runs with `vulnerabilityScan: disable` (the `minimal capabilities` case inherits the `enable` default), so the "secrets rule absent" path is only covered by the description's manual `helm template` — a small `notContains` assertion in `tests/` would lock it in.
- `operator/admission-webhook/role.yaml` still has an unpinned `get` on secrets for the same certgen pattern; pinning it to `kubescape-admission.secretName` would be the consistent follow-up.

PR #928, opened 2026-09-19
fix(image): add maintained kubectl tools image

### LINKED ISSUE #923: `quay.io/kubescape/kubectl:1.36.1` has 25 critical CVEs and its base can no longer be refreshed
# Description

`quay.io/kubescape/kubectl:1.36.1` is used by `certificates.certgen.image` and `grypeOfflineDB.rollout.image`.

It currently contains 25 critical CVEs, all inherited from its Photon OS 5.0 base image.

Simply bumping the tag is not an option. This image is re-tagged from `bitnami/kubectl`, but Bitnami has removed all versioned tags from the public Docker Hub repository. Only `latest` remains, so there is no newer `bitnami/kubectl:<version>` image available to re-tag.

# Environment

Chart: `kubescape-operator` 1.40.4 (current release)  
Image: `quay.io/kubescape/kubectl:1.36.1` (introduced in #843)

# Steps To Reproduce

```bash
trivy image --severity CRITICAL quay.io/kubescape/kubectl:1.36.1
skopeo list-tags docker://docker.io/bitnami/kubectl   # non-digest tags: ["latest", "latest-metadata"]
```

# Expected behavior

The certgen image should use a base image that continues to receive security updates.

# Actual Behavior

The image currently reports 25 critical vulnerabilities:

| package | criticals | installed | fixed in |
| --- | --- | --- | --- |
| `curl`, `curl-libs` | 16 | `8.19.0-2.ph5` | `8.21.0-0.0.1.ph5` |
| `gawk`, `gawk-bin` | 4 | `5.3.2-3.ph5` | `5.3.2-5.ph5` |
| `openssl`, `openssl-libs` | 2 (CVE-2026-34182) | `3.5.6-2.ph5` | `3.5.7-1.ph5` |
| `libssh2` | 2 (CVE-2026-7598, BDSA-2026-9020) | `1.11.1-1.ph5` | `1.11.1-3.ph5` |
| `nss-libs` | 1 (CVE-2026-16389) | `3.78-12.ph5` | `3.78-13.ph5` |

The Photon base is dated `20260214`, and the image itself was built on 2026-05-12. Version 1.40.4 still ships this image.

I do not see an obvious exploit path in the current usage. `certgen-create.sh` and `certgen-patch.sh` do not call `curl`, `gawk`, `libssh2`, or NSS. For the OpenSSL CVE, the script only passes the private key, CSR, and configuration it has just generated.

The vulnerabilities are still present in image scans, though. This affects the storage Deployment in every cluster running the chart. The init container also runs with a ServiceAccount that has `secrets: [get, create]` and `apiservices: [get, patch]`, so leaving a base with known critical CVEs is not ideal.

# Additional context

A replacement image needs to include `bash`, `kubectl`, `jq`, and `openssl`, as well as `base64`, `mktemp`, and `seq`. It also needs to work with an arbitrary non-root UID.

The usual slim alternatives do not meet those requirements:

- `registry.k8s.io/kubectl` and `rancher/kubectl` do not include a shell.
- `alpine/kubectl` is essentially kubectl only.
- `alpine/k8s` includes `libssl3` and `libcrypto3`, but not the `openssl` binary.

There are three possible approaches:

1. Re-tag `bitnami/kubectl:latest`. This is still the same upstream image, but it currently uses Photon base `20260727`, includes kubectl 1.37.0, and had 0 critical CVEs as of 2026-09-15. The downside is that the tag no longer contains a version. Consumers would need to pin the image by digest, version-based tooling becomes less useful, and Renovate reports it as unversioned.

2. Publish and maintain a project-owned image based on Alpine with `kubectl`, `jq`, `openssl`, and `bash`. This removes the dependency on Bitnami's tagging policy and allows the image tag to retain a meaningful version.

3. Remove the `jq` and `openssl` dependencies so the certgen job can use a distroless kubectl image. This would be the largest change, since CA generation would need to move into the operator or into a small Go helper.

I can send a PR for any of these approaches.

ANAMASGARD:
## Overview

Add the source and release workflow for a project-owned `kubectl-tools` image to replace the aging Bitnami-derived image identified in #923.

The image uses a digest-pinned Alpine base and includes the commands required by certificate generation and the Grype offline DB rollout job. The workflow builds amd64 and arm64 images, tests the chart’s UIDs (65532 and 100), scans for critical CVEs before publication, and refuses to overwrite a released tag. A scheduled check opens a PR when the Alpine base digest changes.

This is the first of two PRs. It does not change Helm chart defaults. A separate chart PR will follow maintainer approval, publication, and verification of the public image.

## How to Test

1. Build the image:
   ```sh
   docker buildx build --platform linux/amd64 --load \
     -t kubectl-tools:test -f images/kubectl-tools/Dockerfile .
   ```
2. Run the UID and command checks in `images/kubectl-tools/README.md`.
3. Confirm both architecture jobs pass in the `kubectl-tools image` workflow, including the pre-publication critical-CVE scans.
4. Before publishing, configure the public Quay repository, immutable tags, environment reviewers, and Quay credentials as documented.

Locally, the amd64 build and kubectl checksum verification passed. Required commands worked at UIDs 65532 and 100 with a read-only root filesystem. Trivy reported **0 critical findings** for the amd64 image. `helm lint` passed. Arm64 CI, Quay publication, public pull, and cluster integration tests have not run yet.

## Related issues/PRs

- Part of #923. The follow-up chart PR will replace the default image references after publication.

- [ ] I have commented on my code, particularly in hard-to-understand areas
- [ ] New and existing unit tests pass locally with my changes

Open it against `main`. Use **“Part of #923”** rather than **“Resolved #923”** because the chart still uses the vulnerable image until PR 2.

## Summary by CodeRabbit

- **New Features**
  - Added a `kubectl-tools` container image for amd64 and arm64 platforms.
  - Includes kubectl and supporting utilities for certificate initialization and offline vulnerability database workflows.
  - Added automated image builds, functional tests, vulnerability scanning, and controlled publishing.

- **Documentation**
  - Added usage, release, verification, and maintenance guidance for the container image.

- **Chores**
  - Added weekly checks to refresh and securely pin the Alpine base image.


.github/workflows/kubectl-tools-base-refresh.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+          path = Path('images/kubectl-tools/Dockerfile')
+          source = path.read_text()
+          updated, count = re.subn(r'alpine:3\.24\.1@sha256:[0-9a-f]{64}',
+                                   'alpine:3.24.1@' + sys.argv[1], source)
+          if count != 2:
+              raise SystemExit(f'expected two Alpine references, found {count}')
+          path.write_text(updated)
+          PY
+      - uses: peter-evans/create-pull-request@22a9089034f40e5a961c8808d113e2c98fb63676 # v7
```
matthyx, line 34: **Major — the refresh PR will not get CI.** `peter-evans/create-pull-request` uses `GITHUB_TOKEN` by default, and GitHub deliberately does not fire `pull_request` (or any other) workflows for events created with `GITHUB_TOKEN` (see the action's own docs: *"Workflows triggered by pull requests created by this action will not run"*). So the `kubectl-tools image` build/UID/Trivy jobs the README tells the reviewer to check will not run on the refresh PR until someone closes/reopens it or pushes to it.

The repo already uses `actions/create-github-app-token` in `01-update_tag.yaml`; passing that token (or a PAT) as `token:` here fixes it. Alternatively drop the automatic PR and just fail the scheduled job when the digest drifts, which needs no extra credentials.

images/kubectl-tools/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM alpine:3.24.1@sha256:28bd5fe8b56d1bd048e5babf5b10710ebe0bae67db86916198a6eec434943f8b AS downloader
+
+ARG TARGETARCH
+ARG KUBECTL_VERSION=v1.36.1
```
matthyx: **Major — kubectl is pinned three patch releases behind.** `https://dl.k8s.io/release/stable-1.36.txt` is `v1.36.4` today. Trivy on the image built from this commit:

```
usr/local/bin/kubectl: HIGH=22, MEDIUM=8
  stdlib v1.26.2      → fixed in 1.26.3..1.26.6 (CVE-2026-33811, -33814, -39820, -39821, -39836, -42499, -42504, -27145, -39822, -33818, -56853, -56858, -56859, -56860, -56862, -46600)
  golang.org/x/net v0.49.0 → fixed in 0.53–0.56 (CVE-2026-33814, -25681, -27136, -39821, -46600)
  golang.org/x/text v0.33.0 → fixed in 0.39.0 (CVE-2026-56852)
```

The CI gate is CRITICAL-only so this passes, but a brand-new image whose reason to exist is CVE hygiene should start from the current patch. Suggest `KUBECTL_VERSION=v1.36.4` and `TAG: 1.36.4-r1` (the chart's `1.36.1` tag is only a historical accident of the Bitnami re-tag, nothing depends on that exact patch level).

.github/workflows/kubectl-tools-base-refresh.yaml
```diff
@@ -0,0 +1,42 @@
...
+          sudo apt-get install -y skopeo
+          digest=$(skopeo inspect docker://docker.io/library/alpine:3.24.1 --format '{{.Digest}}')
+          python3 - "$digest" <<'PY'
+          from pathlib import Path
+          import re
+          import sys
+
+          path = Path('images/kubectl-tools/Dockerfile')
+          source = path.read_text()
+          updated, count = re.subn(r'alpine:3\.24\.1@sha256:[0-9a-f]{64}',
```
matthyx: Non-blocking: this only follows re-pushes of the `3.24.1` tag, which Alpine rarely does — security fixes land as new patch tags (`3.24.2` is already out) and in the package repos. Inside the image built from this commit:

```
$ apk upgrade --no-cache --simulate
(1/4) Upgrading alpine-release (3.24.1-r0 -> 3.24.2-r0)
(2/4) Upgrading libapk (3.0.6-r0 -> 3.0.8-r0)
(3/4) Upgrading ca-certificates-bundle (20260611-r0 -> 20260909-r0)
(4/4) Upgrading apk-tools (3.0.6-r0 -> 3.0.8-r0)
```

Two cheap improvements: add `apk upgrade --no-cache` before the `apk add` in both stages (so base packages like `libcrypto3` get fixes at build time), and/or inspect `alpine:3.24` here with a regex like `alpine:3\.24(\.\d+)?@sha256:...` so patch releases are picked up.

matthyx (changes requested):
Thanks for picking up #923 — the image itself is a big improvement over the Photon-based one. I built it locally at `5211587` (amd64): `docker buildx build` OK, all required tools present and working at UIDs 65532 and 100 under `--read-only`, and Trivy reports **0 CRITICAL / 0 OS-package findings** vs 25 CRITICAL + 57 HIGH for `quay.io/kubescape/kubectl:1.36.1`. CI (both arches, Trivy gate) is green and the action SHAs in the last commit all resolve to the tags they claim.

Requesting changes for two concrete issues (inline), plus one thing that needs a maintainer decision before this can go anywhere:

**Blockers / should-fix**
1. The scheduled refresh workflow opens its PR with the default `GITHUB_TOKEN`, so the `kubectl-tools image` workflow will never run on that PR (GitHub does not trigger `pull_request` workflows for events created by `GITHUB_TOKEN`). The README's step 'review its two architecture builds and scans' cannot happen as written. See inline.
2. `KUBECTL_VERSION=v1.36.1` is three patch releases behind `stable-1.36` (`v1.36.4` today). Trivy on the built image shows 22 HIGH in `/usr/local/bin/kubectl` (Go stdlib 1.26.2 + x/net 0.49.0), all fixed in newer builds. For a PR whose purpose is CVE hygiene, a fresh image should not start out behind. See inline.

**Needs maintainer decision (not something the PR can fix)**
- #923 lists three approaches and no maintainer has weighed in yet. This PR commits the project to option 2: building and hosting an image out of `helm-charts` (which has never built images), a new public `quay.io/kubescape/kubectl-tools` repo (doesn't exist yet), a `quay-kubectl-tools` GH environment and Quay secrets. That's an org-level choice; I'd like an explicit ack from the maintainers on (a) option 2 vs. re-tagging `bitnami/kubectl:latest` by digest as a stop-gap, and (b) whether `helm-charts` is the right home for the Dockerfile/release workflow.

**Non-blocking**
- The base refresh only tracks the digest of the fixed `3.24.1` tag; Alpine `3.24.2` is already out and `apk upgrade --simulate` inside the built image lists 4 upgradable packages (`alpine-release`, `libapk`, `apk-tools`, `ca-certificates-bundle`). Adding `apk upgrade --no-cache` to both stages and/or tracking the `3.24` minor tag would make the refresh actually catch Alpine security fixes. See inline.
- `TAG` (workflow) and `KUBECTL_VERSION` (Dockerfile) encode the same version in two files; consider deriving one from the other or passing `KUBECTL_VERSION` as a build-arg from the workflow.
- Branch name `codex/refresh-kubectl-tools-base` — a neutral prefix like `chore/` fits the repo better.
- Heads-up for the follow-up chart PR: `mktemp -d` in `certgen-create.sh` needs a writable `/tmp`; it works today because the certgen init container does not set `readOnlyRootFilesystem`, but if that gets hardened later an `emptyDir` will be needed.

Reviewed head: `521158790d20737f391fd227684d96cb619ea68a` → `main` (`9ab1f2c`).

matthyx:
Superseded by #929.

Thanks for the work here — the build/test/scan/publish workflow structure and the immutable `-rN` tag release process you documented in this PR's README were carried over directly into the new PR's CI workflows, they're a solid pattern.

#929 takes a different approach for the final image: instead of an Alpine-based image with bash + kubectl + jq + openssl, it reimplements the chart's certgen, rollout-restart, and helm-upgrade logic natively in Go (client-go, the Helm SDK, crypto/x509), so the final image can be fully non-debug-distroless (no shell, no package manager at all) on `gcr.io/distroless/static-debian13:nonroot`, matching kubevuln's `build/Dockerfile` pattern. It also folds in the `helm-release-upgrader` replacement (issue #923's other half) into the same image, rather than leaving that for a separate follow-up.

Closing this in favor of #929 — happy to discuss the tradeoffs there if useful.

PR #103, opened 2023-03-05
OpenTelemetry - KS cloud support

amirmalka:
## Overview

Added support for sending OpenTelemetry data to KS cloud.
Otel collector will be enabled by default, with ksCloud exporter enabled by default

PR #930, opened 2026-09-21
test: add Agent Runtime posture deployment smoke test

dakshhhhh16:
## What this changes

This adds a small chart-level smoke suite for the Agent Runtime posture path. The existing RBAC test covers the permissions in isolation. These cases exercise the values together as an operator installation would use them.

The suite checks that a normal installation still leaves Agent Runtime posture disabled, does not select `AgentRuntimeHardening`, and does not add the supported Agent Runtime permissions. With the opt-in values enabled, it checks that the capability reaches the operator configuration, the framework is written into `clusterData`, and the scanner receives read-only access to `Sandbox`, `SandboxTemplate`, and `WorkerPool`. `ActorTemplate` is intentionally excluded because it is not part of the scanner's current evidence contract.

It also verifies that the chart continues to render Kubescape's existing validating admission webhook with this configuration.

## Scope note

The chart does not render CEL `ValidatingAdmissionPolicy` resources. Those are deployed by Kubescape's VAP workflow after the policy library is available. This PR only covers the Helm resources the chart owns, so it does not imply that the chart deploys the CEL policy set.

## Validation

- Parsed `agent_runtime_posture_deployment_test.yaml` with Ruby YAML.
- Ran `git diff --check`.
- Attempted the repository's Docker-based `helmunittest/helm-unittest` command. It could not run because the local Docker daemon is unavailable. GitHub Actions will run the chart unit-test job for this PR.

## Summary by CodeRabbit

- **Changes**
  - Agent Runtime posture access now includes read-only permissions for Sandbox, SandboxTemplate, and WorkerPool resources when enabled.
  - Access to Actortemplate resources remains excluded.
  - Agent Runtime posture remains disabled by default, preserving existing scan selection and permissions.

- **Tests**
  - Added Helm coverage for disabled and enabled posture configurations, including framework selection, required access, and admission webhook deployment.

- **Documentation**
  - Updated configuration guidance to reflect supported Agent Runtime resources and default behavior.


matthyx (changes requested):
Request changes — reviewed `f3a586ed267ee5698902de51eb7864e4d6683cc0` against `main` (`9ab1f2c85d25e6fe7a2215fe792536b52ae43633`).

**Blocking — `charts/kubescape-operator/tests/agent_runtime_posture_deployment_test.yaml:48` and `:103`: do not make `ate.dev/actortemplates` a required Agent Runtime permission.**

The assertions reproduce the rule introduced in #893, but the current `AgentRuntimeHardening` contract supports only `Sandbox`, `SandboxTemplate`, and `WorkerPool`. The later framework implementation explicitly excludes `ActorTemplate`: it is an ATE protobuf resource rather than a Kubernetes CRD, so the scanner has no supported evidence source ([regolibrary#761](https://github.com/kubescape/regolibrary/pull/761)); the matching admission-library correction was also merged in [#103](https://github.com/kubescape/cel-admission-library/pull/103). Preserving this stale grant as a smoke-test expectation makes the least-privilege correction fail in CI and keeps an unnecessary read permission on enabled installations.

Please update the chart RBAC contract and these default/enabled assertions to the current three-resource scope (or split the production RBAC correction into its own PR and make this suite depend on it). The pre-existing focused RBAC suite must be aligned as well.

The rest of the one-file test addition is useful combined-value coverage. `defaultFrameworks` default omission is consistent with merged #890, and no duplicate deployment suite was found. Local validation: `git diff --check`; `helm unittest charts/kubescape-operator` (16 suites, 118 tests, 1060 snapshots); `helm lint charts/kubescape-operator`; all passed. GitHub checks are also green, but they do not address this obsolete-contract regression.

matthyx (approved):
Re-reviewed `60b44d91631683f2ec52fe07638583f1aa0e43b5` against `main` (`9ab1f2c85d25e6fe7a2215fe792536b52ae43633`). The requested least-privilege correction is complete: enabled posture now grants read-only access only to Sandbox, SandboxTemplate, and WorkerPool; `ActorTemplate` is explicitly excluded in both focused and smoke tests. The values/README text matches the current framework contract. Local validation passed: `git diff --check`, `helm unittest charts/kubescape-operator` (16 suites, 118 tests, 1060 snapshots), `helm lint charts/kubescape-operator`, and a rendered enabled chart. Current GitHub checks are green. Approved.

